PR #14, opened 2019-10-11
Blocks delete bug

nanetteCodes:
The bug was associated to sweet alerts.  The API has been updated. "swal" is now an object which has a method "fire".


efjacobson:
@nanetteCodes approved, just  need  to resolve conflict

PR #21, opened 2019-10-16
Slug double dash

efjacobson:
https://github.com/triniti/cms-js/issues/20

### DIFF
--- a/src/plugins/ncr/components/slug-editor/index.jsx
+++ b/src/plugins/ncr/components/slug-editor/index.jsx
@@ -99,7 +99,12 @@ class SlugEditor extends React.Component {
 
   handleChange({ target: { value: slug } }) {
     this.setState((prevState, { initialSlug }) => ({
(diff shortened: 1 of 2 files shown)


efjacobson:
@arunkarnati good suggestions, implemented

src/plugins/ncr/components/slug-editor/index.jsx
```diff
@@ -99,7 +99,13 @@ class SlugEditor extends React.Component {
 
   handleChange({ target: { value: slug } }) {
     this.setState((prevState, { initialSlug }) => ({
-      slug: slug.replace(' ', '-'),
+      slug: slug
+        .replace(/^[-/\s]|[\s]$/g, '')
```
arunkarnati: this looks good, thanks for adding these, one really tiny change and we are done:

```suggestion
        .replace(/^[-/\s]+|[\s]+$/g, '')
```

allows catching things like `/ /2019/10/16`

Issue #17, opened 2019-10-15
Cannot upload images


nanetteCodes:
PR: https://github.com/triniti/cms-js/pull/26

PR #26, opened 2019-10-16
changed baseUrls.image from s to zed

nanetteCodes:
issue: https://github.com/triniti/cms-js/issues/17

### DIFF
--- a/src/plugins/dam/utils/damUrl.js
+++ b/src/plugins/dam/utils/damUrl.js
@@ -1,16 +1,11 @@
-/* globals APP_ENV, API_ENDPOINT */
+/* globals DAM_BASE_URL, IMAGE_BASE_URL */
 import AssetId from '@triniti/schemas/triniti/dam/AssetId';
 import Message from '@gdbots/pbj/Message';
 import NodeRef from '@gdbots/schemas/gdbots/ncr/NodeRef';
 
-const tld = API_ENDPOINT.split('://').pop().split('/').shift()
-  .split('.')
-  .slice(-2)
-  .join('.');
-
 export const baseUrls = {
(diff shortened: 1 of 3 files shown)


src/plugins/dam/utils/damUrl.js
```diff
@@ -10,7 +10,7 @@ const tld = API_ENDPOINT.split('://').pop().split('/').shift()
 
 export const baseUrls = {
   dfault: `https://dam.${APP_ENV === 'prod' ? '' : `${APP_ENV}.`}${tld}/`,
-  image: `https://images${APP_ENV === 'prod' ? '.' : `-${APP_ENV}.`}${tld}/`,
+  image: `https://imagez${APP_ENV === 'prod' ? '.' : `-${APP_ENV}.`}${tld}/`,
```
albertgcsula: @gdbrown due to the differences in the image base urls i.e. https://imagez-dev.toofab.com vs. https://bachelornation-images-dev.akamaized.net. Can we just set something like: IMAGE_BASE_URL=https://imagez-dev.toofab.com/ in the env file and use image:IMAGE_BASE_URL here?

gdbrown: @albertgcsula not opposed to that, was trying to get them to be the same strategy as the symfony apps and let the index.js or config on the local site to customize it as needed.  either way.  if using globals then you have some more wiring to do in other places and you'd want to do DAM_BASE_URL and IMAGE_BASE_URL

albertgcsula:
@efjacobson asking for your review on my changes, thanks.

efjacobson:
looks good to me. @gdbrown i wasnt able to find any other places that need wiring

gdbrown:
@efjacobson when you see it, you'll know.  (it's at the site level)

PR #32, opened 2019-10-17
Asset tags

efjacobson:


src/plugins/dam/components/image-asset-fields/index.jsx
```diff
@@ -13,65 +13,68 @@ import TextareaField from '@triniti/cms/components/textarea-field';
...
-    </CardBody>
-  </Card>,
-  schemas.node.hasMixin('triniti:apollo:mixin:has-poll') && (
-    <Card key="poll">
-      <CardHeader>Poll</CardHeader>
-      <CardBody>
-        <FieldArray
-          component={PollPickerField}
+const ImageAssetFields = ({ asset, isEditMode, schemas }) => (
+  <>
```
arunkarnati: @efjacobson - ha, this is new to me, empty tags instead of fragments?

efjacobson: @arunkarnati yep https://reactjs.org/docs/fragments.html#short-syntax

PR #48, opened 2019-10-21
add fix for widget redux form fields

efjacobson:


src/plugins/curator/screens/widget/Form.jsx
```diff
@@ -18,44 +18,65 @@ import schemas from './schemas';
  *
  * @param {string} type
  */
+let fieldsComponents = {};
 const getFieldsComponent = (type) => {
+  if (fieldsComponents[type]) {
+    return fieldsComponents[type];
+  }
   switch (type) {
```
arunkarnati: @efjacobson - can we shorten the code using template literal? something like this:
```suggestion
  fieldsComponents[type] = createLazyComponent(import(`@triniti/cms/plugins/curator/components/${type}-fields`));
```

arunkarnati: @efjacobson - interesting, with that said, should we update all the other places where we used `dynamic import` syntax with dynamic component resolution? eg: https://github.com/wb-apps/tmz-cms/blob/master/packages/triniti-cms-js/plugins/curator/components/create-teaser-modal/Form.jsx#L19

efjacobson: @arunkarnati yeah, probably. can have qa make a ticket

PR #59, opened 2019-10-24
added label for related galleries

cohencodes:


src/plugins/curator/components/gallery-fields/index.jsx
```diff
@@ -92,6 +92,7 @@ const GalleryFields = ({ isEditMode, gallery, formName, nodeRef, schemas }) => {
           name="allowComments"
         />
         <hr />
+        <Label style={{ display: 'block' }}>Related Galleries</Label>
```
arunkarnati: @cohencodes - why add here when you can just pass a `label` prop to the `GalleryPickerField`? see here: https://github.com/triniti/cms-js/blob/master/src/plugins/curator/components/gallery-picker-field/index.jsx#L24

update the code in the `picker-field` to show the label in both view & edit modes.

efjacobson: @cohencodes instead of doing an inline style, i encourage you to compare the working version and this version to see why they are different. you'll see that in the working version the picker component is wrapped in a FormGroup. so do that instead, make sure the other places where the galleryPickerField are used are not adversely affected, and do that instead

efjacobson: @arunkarnati it has two labels, one inside ("SEARCH AND SELECT GALLERIES") and one outside ("RELATED GALLERIES")

arunkarnati:
@efjacobson @cohencodes - spoiler alert, it might start showing dupes once you add a related gallery, https://github.com/triniti/cms-js/blob/master/src/plugins/curator/components/gallery-picker-field/SortableList/index.jsx#L74

efjacobson:
lol good catch @arunkarnati . @cohencodes can you verify?

cohencodes:
Screen Shot 2019-10-24 at 12 37 22 PM
looks good to me

efjacobson:
@cohencodes save it and enter view mode:

Screen Shot 2019-10-24 at 12 40 57 PM

PR #60, opened 2019-10-24
Prefill notification modal

efjacobson:


src/plugins/notify/components/create-notification-modal/Form.jsx
```diff
@@ -47,6 +47,7 @@ const Form = ({ contentChangeable, type, apps }) => (
             return (
               <FieldArray
                 component={ArticlePickerField}
+                isEditMode={contentChangeable}
```
hunyoboy: Cool, good to know I can pass EditMode on FieldArray.

PR #61, opened 2019-10-24
add related gallery title when in view mode and no results

arunkarnati:


cohencodes:
So It's different from what we have now, just a matter of preference on what you guys think is better:

Current Version:

Screen Shot 2019-10-24 at 1 20 34 PM


Arun's Changes:

Screen Shot 2019-10-24 at 1 20 34 PM

cohencodes:
@arunkarnati also in Structure>Teasers>Galleries it now says "Related Galleries" instead of Selected Galleries.

cohencodes:
@arunkarnati I think it should say Search and Select Galleries like the current version imo

arunkarnati:
@cohencodes - teasers label is fine, but the problem with the code that we have in `release-next` is: it dupes the `label` when in `view-mode`, here's a screenshot:

image

PR #82, opened 2019-10-28
History revisions

bhsiao55:


src/plugins/pbjx/components/event-stream/index.jsx
```diff
@@ -59,10 +61,18 @@ class EventStream extends React.Component {
     if (events.length) {
       if (!prevProps.events.length || (prevProps.events[0].get('event_id') !== events[0].get('event_id'))) {
         this.setState({ allEvents: [...allEvents, ...events] });
+        this.setState({ loadMore: false });
```
hunyoboy: i think you can combine this in one setState

bhsiao55: modified

PR #102, opened 2019-11-01
13 dam batch edit

rsumilang:


src/plugins/dam/components/batch-edit-modal/selector.js
```diff
@@ -0,0 +1,54 @@
...
+
+import { formNames } from '../../constants';
+
+/**
+ * @param {Object} state     - The entire redux state.
+ * @returns {Object}
+ */
+export default (state) => {
+  const formName = `${formNames.BATCH_EDIT}`;
+  const { isBatchEditOpen } = state.dam.batchEdit;
```
efjacobson: @rsumilang use selectors to get stuff from the store

src/plugins/dam/actions/displayBatchEdit.js
```diff
@@ -0,0 +1,6 @@
+import { actionTypes } from '@triniti/cms/plugins/dam/constants';
+
+export default (display) => ({
+  type: actionTypes.DISPLAY_BATCH_EDIT_REQUESTED,
```
efjacobson: @rsumilang why is this whole thing in redux? couldnt you handle this with component state in the button?

src/plugins/curator/components/sortable-grid/ButtonToolbarAsset.jsx
```diff
@@ -65,12 +74,22 @@ export default class ButtonToolbarAsset extends React.Component {
...
+          </ButtonToolbar>
+        )}
         {!disabled && isHovering && (
-          <ButtonToolbar style={{ position: 'absolute', right: '.5rem', top: '.5rem' }}>
+          <ButtonToolbar style={{ position: 'absolute', left: '0', top: '0', padding: '.5rem', width: '100%', justifyContent: 'space-between', minHeight: '44px' }}>
             {!showEditSequence && (
               <>
-                <ToolbarButton icon="edit" id="media-edit" tooltip="Edit" onMouseDown={() => onEditAsset(node)} />
-                <ToolbarButton icon="trash" id="media-remove" tooltip="Remove" onMouseDown={() => onRemoveAsset(node)} />
+                {multiSelect && <Checkbox checked={isSelected} onChange={() => onSelect(node)} />}
```
efjacobson: @rsumilang is multiSelect ever false? is this prop  needed?

src/plugins/curator/components/gallery-media/index.jsx
```diff
@@ -273,12 +278,25 @@ class GalleryMedia extends React.Component {
 
   handleChangeSearchParam(key, value) {
     const { delegate, searchNodesRequestState: { request } } = this.props;
-    const newRequest = { ...request.toObject(), [key]: value };
+    const newRequest = Object.assign({}, request.toObject(), {
```
efjacobson: @rsumilang spread wont work here?

src/plugins/curator/components/gallery-media/index.jsx
```diff
@@ -324,7 +343,16 @@ class GalleryMedia extends React.Component {
             onIncreaseImagesPerRow={this.handleIncreaseImagesPerRow}
             onSlideImagesPerRow={this.handleSlideImagesPerRow}
           />
-          <Button disabled={!isEditMode} onClick={this.handleToggleModal} className="mr-0 mt-2 mb-2">Add Images</Button>
+          <div className="d-inline-flex flex-wrap justify-content-end ml-2 my-1">
+            <BatchEditButton
+              assetIds={selected}
+              className="my-1 mr-0"
+              isEditMode={isEditMode}
+              node={getNode(nodeRef)}
```
efjacobson: @rsumilang is this being used somewhere? i dont see it in the button

src/plugins/curator/components/sortable-grid/ButtonToolbarAsset.jsx
rsumilang: There is nothing else that currently uses SortableGrid. However the component seems generic enough to display assets anywhere with a toolbar.  However I can imagine a thing that may want to display assets in the future in a grid but not require any functionality to individually select assets so I gave it an option.

src/plugins/dam/components/batch-edit-modal/delegate.js
```diff
@@ -0,0 +1,87 @@
+import { reset } from 'redux-form';
+
+import AbstractDelegate from '@triniti/cms/plugins/ncr/screens/node/AbstractDelegate';
+import batchEditPatchAssets from '@triniti/cms/plugins/dam/actions/batchEditPatchAssets';
+import displayBatchEdit from '@triniti/cms/plugins/dam/actions/displayBatchEdit';
+
+import schemas from './schemas';
+import { formNames } from '../../constants';
+
+class Delegate extends AbstractDelegate {
```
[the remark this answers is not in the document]
rsumilang: this has been updated to not use the abstract delegate. this came out to about the same size.

rsumilang:
@efjacobson please review again

src/plugins/dam/actions/batchEditPatchAssets.js
```diff
@@ -0,0 +1,21 @@
+import NodeRef from '@gdbots/schemas/gdbots/ncr/NodeRef';
+import { actionTypes } from '@triniti/cms/plugins/dam/constants';
+
+export default ({ fields, values }, files, config) => {
+  const { schemas } = config;
```
efjacobson: @rsumilang can `schemas.patchAssets` ever be something other than `PatchAssetsV1Mixin.findOne()`?. if not, might make sense to just import that directly instead of passing a config to this action

src/plugins/dam/actions/displayBatchEdit.js
efjacobson: @rsumilang this is not being used?

src/plugins/dam/components/batch-edit-button/index.jsx
```diff
@@ -0,0 +1,69 @@
...
+      ...btnProps
+    } = this.props;
+
+    const {
+      isBatchEditOpen,
+    } = this.state;
+
+    return ([
+      <Button
+        key="a"
```
efjacobson: @rsumilang if you use a fragment for these instead of an array you can avoid putting keys on things

src/plugins/dam/components/batch-edit-modal/BatchEdit.jsx
```diff
@@ -0,0 +1,193 @@
+// TODO: reload the gallery-media component after the batch edit process is complete
```
efjacobson: @rsumilang you can do this pretty easily by keying it to the command's id. will have to re-bind to delegate on update

src/plugins/dam/components/batch-edit-modal/BatchEdit.jsx
```diff
@@ -0,0 +1,193 @@
...
+  handleUpdate() {
+    const {
+      assetIds,
+      currentValues,
+      delegate,
+      onToggleBatchEdit,
+    } = this.props;
+
+    // Slug is in currentValues sometimes so we must have a white list.
+    const filter = ['title', 'credit', 'expiresAt', 'description'];
```
efjacobson: @rsumilang it sometimes has slug because of the sluggable subscriber's onTitleChanged. you could check there to see if the form even has a slug field registered to avoid this workaround. the subscriber's method shouldnt really be adding a slug value for this anyways

src/plugins/dam/components/batch-edit-modal/BatchEditFields.jsx
```diff
@@ -0,0 +1,91 @@
...
+    currentValues: {},
+    onBatchUpdate: noop,
+  };
+
+  constructor(props) {
+    super(props);
+    this.handleBatchUpdate = this.handleBatchUpdate.bind(this);
+  }
+
+  handleBatchUpdate() {
```
efjacobson: @rsumilang when does this get called? far as i can tell the juice is still the button's onClick

src/plugins/dam/components/batch-edit-modal/BatchEdit.jsx
```diff
@@ -0,0 +1,193 @@
...
+    } = this.props;
+
+    return (
+      <Modal isOpen={isOpen} toggle={this.handleToggleUploader} size="md">
+        <ModalHeader toggle={this.handleToggleUploader}>
+          Batch Edit
+        </ModalHeader>
+        <ModalBody>
+          <Card>
+            <Form
```
efjacobson: @rsumilang far as i can tell the only props this really needs is `form` and `initialValues` (and initialValues is a stretch because the form is not getting hydrated, even if only a single asset is selected). also no validation is happening? this is kind of one foot in redux form and one foot in regular react - for example this doesnt need `onSave` or `onSubmit` because the form is never being submitted, its just a click handler on the button

src/plugins/dam/components/batch-edit-modal/delegate.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  getInitialValues() {
+    // can be filled in later with the common values between each node
+    return {};
+  }
+
+  handleValidate(values, formProps) {
+    const formEvent = this.createFormEvent(values, formProps);
+    this.pbjx.trigger(formEvent.getMessage(), SUFFIX_VALIDATE_FORM, formEvent);
```
efjacobson: @rsumilang i dont think any subscribers are actually picking this up. if you paste a mega long string into the title field, you dont get any warnings and can still press the button but that then logs an error in the guard

src/plugins/dam/components/batch-edit-modal/selector.js
```diff
@@ -0,0 +1,52 @@
...
+export default (state) => {
+  const formName = `${formNames.BATCH_EDIT}`;
+  const isFormDirty = isDirty(formName)(state);
+  const isFormPristine = isPristine(formName)(state);
+  const isFormValid = isValid(formName)(state);
+  const initialValues = getFormInitialValues(formName)(state);
+  const currentValues = getFormValues(formName)(state);
+
+  const alerts = getAlerts(state, formName);
+  const enableExpirationDateApplyAll = (() => {
```
efjacobson: @rsumilang what is this for? doesnt it always apply all?

src/plugins/dam/components/batch-edit-modal/BatchEdit.jsx
```diff
@@ -0,0 +1,193 @@
...
+  render() {
+    const {
+      assetIds,
+      nodeRef,
+      delegate,
+      currentValues,
+      getNode,
+      isFormDirty,
+      isFormValid,
+      isFormPrestine,
```
efjacobson: @rsumilang typo

src/plugins/dam/components/batch-edit-modal/selector.js
```diff
@@ -0,0 +1,52 @@
...
+
+    if (initialValues.expiresAt) {
+      return !areDatesEqual(initialValues.expiresAt, currentValues.expiresAt);
+    }
+
+    return true;
+  })();
+
+  return {
+    alerts,
```
efjacobson: @rsumilang what is this for?

src/plugins/dam/constants.js
```diff
@@ -26,6 +27,7 @@ export const actionTypes = {
   PREFIX: PLUGIN_PREFIX,
 
   PATCH_ASSETS_REQUESTED: t('PATCH_ASSETS_REQUESTED'),
+  DISPLAY_BATCH_EDIT_REQUESTED: t('DISPLAY_BATCH_EDIT_REQUESTED'),
```
efjacobson: @rsumilang the action that uses this is unnecessary so it think this is too

src/plugins/dam/actions/batchEditPatchAssets.js
rsumilang: I think it's always going to be `PatchAssetsV1Mixin.findOne()`. If I import that directly there then the action looks a little funny because it's just passing `config` straight through without getting anything from it. That of course is totally fine but if I already have what I need in the config then I think it makes sense to get it from there instead of importing it running `findOne()` again to get the value of something I already have mainly because I can't remove the `config` object, the saga is going to need additional things that the schemas object provides.

efjacobson: @rsumilang up to you i guess...not sure why you have to pass in the config though. couldnt you just do something like:

```
import NodeRef from '@gdbots/schemas/gdbots/ncr/NodeRef';
import { actionTypes } from '@triniti/cms/plugins/dam/constants';
import AssetPatchedV1Mixin from '@triniti/schemas/triniti/dam/mixin/asset-patched/AssetPatchedV1Mixin';
import PatchAssetsV1Mixin from '@triniti/schemas/triniti/dam/mixin/patch-assets/PatchAssetsV1Mixin';

export default ({ fields, values }, files) => {
  const nodeRefs = files.map((nodeRefStr) => NodeRef.fromString(nodeRefStr));

  const data = {
    paths: fields,
    ...values,
  };

  const pbj = PatchAssetsV1Mixin.findOne().createMessage(data);
  pbj.addToSet('node_refs', nodeRefs);

  const config = {
    schemas: {
      assetPatched: AssetPatchedV1Mixin.findOne(),
    },
  };

  return {
    type: actionTypes.PATCH_ASSETS_REQUESTED,
    config,
    pbj,
  };
};
```

and not include it in either the uploader or batch-edit schemas? the way it is now it is being done twice. since this action only ever does one thing it could go here, unlike other actions which might do update-artice and update-video and therefore need a config

rsumilang: If I do this here then I will have to change this in the delegate to import PatchAssetsV1Mixin as well for the creating form function. At that point it starts to feel like leaving the schemas file alone is fine because more than one thing is referencing it and to change it is only in one familiar location (schemas.js).

efjacobson:
over to you @gdbrown

PR #106, opened 2019-11-06
#1043 - Persistent Search Filters On Remount

hunyoboy:


src/plugins/curator/screens/search-teasers/delegate.js
```diff
@@ -10,9 +10,13 @@ class Delegate extends AbstractDelegate {
   }
 
   getSearchParams() {
-    const { statuses } = this.component.props;
+    const { searchNodesRequestState: { request }, statuses } = this.component.props;
 
-    return { sort: SearchTeasersSort.ORDER_DATE_DESC.getValue(), statuses };
+    return {
+      sort: SearchTeasersSort.ORDER_DATE_DESC.getValue(),
```
gdbrown: @hunyoboy shouldn't sort come from the request too if request is populated?

src/plugins/dam/screens/search-assets/delegate.js
```diff
@@ -7,6 +7,15 @@ class Delegate extends AbstractDelegate {
       schemas,
     }, dependencies);
   }
+
+  getSearchParams() {
+    const { searchNodesRequestState: { request }, statuses } = this.component.props;
+
+    return {
+      statuses,
```
gdbrown: @hunyoboy what about sort here?

src/plugins/curator/screens/search-teasers/delegate.js
hunyoboy: i agree.. thank you

src/plugins/dam/screens/search-assets/delegate.js
hunyoboy: @gdbrown i added the sort param

gdbrown: @hunyoboy don't destruct enums, they are not plain ole objects

hunyoboy:
> @hunyoboy are we certain we've covered all the search screens and the props/params that apply on each one?

As far as I know, yes.

hunyoboy:
@gdbrown i believe i covered all the search forms that needed to be persistent. but please let me know if i miss something. thanks.

PR #120, opened 2019-11-08
Fixed broken thumbnail in asset uploader.

jalopez13:
Issue: https://github.com/wb-apps/tmz-cms/issues/1079

Fixed broken thumbnail in asset uploader.

### DIFF
--- a/src/plugins/dam/components/asset-snippet/Image.jsx
+++ b/src/plugins/dam/components/asset-snippet/Image.jsx
@@ -23,8 +23,7 @@ const Image = ({ asset, previewUrl }) => (
       </FormGroup>
     </div>
     <div className="col-12 col-md-6">
-      {console.log('Asset: ', JSON.parse(asset))}
-      {!previewUrl && <div className="spinner-wrapper"> <Spinner className="m-0 mr-2" color="#08a0e8" width="24" strokeWidth="8" /> </div>}
+      {!previewUrl && <Spinner centered />}
 
       {previewUrl && (
       <div className="dam-media-container">
--- a/src/plugins/dam/components/asset-snippet/Image.jsx
+++ b/src/plugins/dam/components/asset-snippet/Image.jsx
@@ -23,6 +23,7 @@ const Image = ({ asset, previewUrl }) => (
       </FormGroup>
     </div>
     <div className="col-12 col-md-6">
+      {previewUrl && (
       <div className="dam-media-container">
         <ReactImageMagnify
           smallImage={{
@@ -63,6 +64,7 @@ const Image = ({ asset, previewUrl }) => (
           enlargedImageClassName="h-100 w-auto asset-hover-mw"
         />
       </div>
+      )}
     </div>
   </div>
 );
--- a/src/plugins/dam/components/asset-snippet/Image.jsx
+++ b/src/plugins/dam/components/asset-snippet/Image.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import ReactImageMagnify from 'react-image-magnify';
 import Message from '@gdbots/pbj/Message';
-import { FormGroup, Label } from '@triniti/admin-ui-plugin/components';
+import { FormGroup, Label, Spinner } from '@triniti/admin-ui-plugin/components';
 import damUrl from '../../utils/damUrl';
 
 const Image = ({ asset, previewUrl }) => (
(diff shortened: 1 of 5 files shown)


src/plugins/dam/sagas/uploadFileFlow.js
```diff
@@ -30,9 +30,9 @@ export default function* (hashName, fileInfo, fileBuffer, preSignedUrl, variant)
 
   for (let retries = 0; retries < MAX_RETRIES; retries += 1) {
     try {
-      // Delay the process by 2 seconds if this isn't the first try
+      // Delay the process by 4 seconds if this isn't the first try
       if (retries) {
-        yield delay(2000);
+        yield delay(4000);
```
gdbrown: @jalopez13 are you certain this is the fix?  this is for the upload process, not hte preview part correct?  i thought the broken thumb was the previewing of the asset and the upload was in fact still working.

seems like we may want to do the `yield call(checkImage, previewUrl)` multiple times before assuming a re-upload is needed.  also, that yield checkImage is currently wrong, it should be yield call(fun, args).

@rsumilang @efjacobson can you also check this out.

jalopez13: @gdbrown 

I updated the delay back to 2 secs. It really makes no difference on the rendering of the thumbnail. Issue was that the component would render but the previewUrl was still not available hence the 403 error I was seeing. So I updated the component to render the thumbnail only when the previewUrl was ready. I tested this over and over and I always see a rendered thumbnail. Do you want me to make any other updates?

src/plugins/dam/components/asset-snippet/Image.jsx
```diff
@@ -23,17 +23,21 @@ const Image = ({ asset, previewUrl }) => (
       </FormGroup>
     </div>
     <div className="col-12 col-md-6">
+      {console.log('Asset: ', JSON.parse(asset))}
```
gdbrown: @jalopez13 remove console.log from component

src/plugins/dam/components/asset-snippet/Image.jsx
```diff
@@ -23,17 +23,21 @@ const Image = ({ asset, previewUrl }) => (
...
+      {!previewUrl && <div className="spinner-wrapper"> <Spinner className="m-0 mr-2" color="#08a0e8" width="24" strokeWidth="8" /> </div>}
```
gdbrown: @jalopez13 are custom styles on spinner really needed here?  those are global styles.  i would use spinner with virtually no customizations if possible, just use the builtin admin-ui or bootstrap classes

src/plugins/dam/components/image-asset-fields/style.scss
```diff
@@ -5,3 +5,15 @@
 .dam-media-container > div > img {
     max-width: 100%;
 }
+
+.spinner-wrapper {
```
gdbrown: @jalopez13 custom css, especially this generic should be avoided if possible.  use the admin-ui components and/or bootstrap utility classes.

src/plugins/dam/components/asset-snippet/Image.jsx
jalopez13: Removed!

src/plugins/dam/components/asset-snippet/Image.jsx
jalopez13: Removed custom styling.

src/plugins/dam/components/image-asset-fields/style.scss
jalopez13: Removed!

src/plugins/dam/sagas/uploadFileFlow.js
jalopez13: Reverted to 2000!

src/plugins/dam/components/asset-snippet/Image.jsx
```diff
@@ -23,17 +23,20 @@ const Image = ({ asset, previewUrl }) => (
       </FormGroup>
     </div>
     <div className="col-12 col-md-6">
+      {!previewUrl && <Spinner centered className="m-0 mr-2" color="#08a0e8" width="24" strokeWidth="8" />}
```
gdbrown: @jalopez13 remove custom color here.  if a style is needed it should come from bootstrap or admin-ui.  primary, secondary, warning, danger, etc. vs a hard coded color.

PR #125, opened 2019-11-14
v0.2.10

efjacobson:
* User can see stale article data


src/plugins/news/services/ArticleSubscriber.js
```diff
@@ -312,14 +317,47 @@ export default class ArticleSubscriber extends EventSubscriber {
     });
   }
 
+  /**
```
gdbrown: @efjacobson doesn't the node screen already pull a fresh copy regardless?  it should.  any time that screen mounts it should be a fresh request.  that would eliminate this need?

efjacobson: @gdbrown right now it does not https://github.com/triniti/cms-js/blob/master/src/plugins/ncr/screens/node/AbstractDelegate.js#L161

PR #127, opened 2019-11-19
1094 instagram block not allowing ids with dash

jalopez13:
Updated regex for instagram block embed url to allow  dashes.


efjacobson:
@jalopez13 pasting in the blockquote from the second to last test crashes the page with tih

Screen Shot 2019-11-19 at 3 38 33 PM

lets get that fixed up and also make the tests more robust so they catch this. you may need to do a parser or something so both regexes can be tested in tandem. take a look at this thing i started and Alon has been working on:

https://github.com/triniti/cms-js/blob/57dc6e2f184f987186f56a12cc4abecc03691ede/src/plugins/blocksmith/components/spotify-embed-block-modal/parseSpotify.js

and

https://github.com/triniti/cms-js/blob/57dc6e2f184f987186f56a12cc4abecc03691ede/src/plugins/blocksmith/components/spotify-embed-block-modal/parseSpotify.test.js

in this way we can test that not only is the id recognized but that it is also properly extracted from the pasted string

PR #131, opened 2019-11-21
made picker field search query persistent

nanetteCodes:


src/plugins/ncr/components/node-picker-field/index.jsx
```diff
@@ -157,20 +158,15 @@ class NodePickerField extends React.Component {
...
-        && (
-          q !== ''
-          || (response && ( // just selecting an item counts as a change, so check for that here
-            response.get('ctx_request').get('q') !== ''
-            && response.get('ctx_request').get('q') !== null
-          ))
-        )
-      ) {
-        this.setState(() => ({ hasStoredFirstSet: false, options: [] }), () => handleSearch(q));
+      if (!isGetAll && action !== 'input-change' && q !== '') {
```
efjacobson: @nanetteCodes just return early from the method itself if the action is not input-change. dont even get into the setState

efjacobson: or i guess @cohencodes

PR #146, opened 2019-12-14
User picker for author_ref

efjacobson:


efjacobson:
waiting for core to get updated

PR #148, opened 2019-12-16
Remove aside from blocks

jalopez13:
Removed aside from blocks:

code-block
divider-block
heading-block
iframe-block
page-break-block
quote-block
text-block


jalopez13:
@efjacobson 

I don't see any non-aside work on this PR. Also the changelog just has a new entry for this PR from current release-next.  I don't see any issues, could you be specific?

PR #151, opened 2020-01-06
add prompt before stopping medialive channel

efjacobson:


nanetteCodes:
@efjacobson looks good

PR #162, opened 2020-01-09
Fix Unit Test's Unresolved Module Error

hunyoboy:
Addressing an issue where disabling a unit test code is necessary due to unresolve module config file error.


hunyoboy:
@efjacobson at your convenience.... Thank you.

efjacobson:
Joel convinced me its better to bring in the dep and not alter the app code, as opposed to import fn workaround in actual component

PR #169, opened 2020-01-11
SpotifyEmbedBlock

cohencodes:
* Add Spotify URI to text area of spotify embed block to make it more clear which content type is being used. 
* Spotify Embed Block requires pasting of embeds twice.


efjacobson:
@cohencodes looks good

PR #170, opened 2020-01-13
v0.3.2

cohencodes:
* Add Spotify URI to text area of spotify embed block to make it more clear which content type is being used. 
* Spotify Embed Block requires pasting of embeds twice.
* YouTube Playlist block: If user manually types anything into input field for EMBED CODE, URL, OR ID the CMS will crash
* Text Editor toolbar does not appear after selecting option, leaving tab, and returning


src/plugins/blocksmith/components/imgur-post-block-modal/getImgurPostBlockId.js
```diff
@@ -1,7 +1,7 @@
 const getImgurPostBlockId = (str) => {
-  const IMGUR_POST_BLOCK_HAS_TYPE = /\/(a|gallery)+\//g;
-  const IMGUR_POST_BLOCK_EMBED_REGEX = /data-id="(a|gallery)?\/?\w+/g;
-  const IMGUR_POST_BLOCK_URL_REGEX = /http?s:\/\/imgur\.com\/(a|gallery)?(\w+)(\/?)\w+(\/?)/g;
+  const IMGUR_POST_BLOCK_HAS_TYPE = /\/(a|gallery)+\//;
+  const IMGUR_POST_BLOCK_EMBED_REGEX = /data-id="(a|gallery)?\/?\w+/;
```
gdbrown: @cohencodes these don't need to be redefined every function call.  move them out of hte function

cohencodes: @gdbrown done

PR #173, opened 2020-01-14
Fix blank slug on enter

efjacobson:


src/plugins/ncr/services/SluggableSubscriber.js
```diff
@@ -37,13 +37,7 @@ export default class SluggableSubscriber extends EventSubscriber {
...
     const node = formEvent.getMessage();
 
-    let slug;
-    if (actualSlug) {
-      slug = actualSlug;
-    } else {
-      slug = createSlug(title);
-    }
-
+    const slug = createSlug(actualSlug || title);
```
hunyoboy: Nice fix and refactor!

PR #176, opened 2020-01-16
Text block hover disappear

efjacobson:


hunyoboy:
@efjacobson Looks good. I cannot replicate all bug.

PR #183, opened 2020-01-24
rearrange download button order

cohencodes:


efjacobson:
@cohencodes looks good to me. @gdbrown what do you think?

src/plugins/ncr/screens/node/index.jsx
```diff
@@ -178,6 +182,14 @@ export default class AbstractNodeScreen extends React.Component {
     this.setState(({ isSaveDropDownOpen }) => ({ isSaveDropDownOpen: !isSaveDropDownOpen }));
   }
 
+  renderAdditionalPrimaryActions() {
```
gdbrown: @cohencodes seems like this extra logic is not needed.  if this just returns `</>` by default you only need override this method and don't need renderAdditionalPrimaryActions and getAdditionalPrimaryActions

PR #185, opened 2020-01-31
arrow down block bug

nanetteCodes:


efjacobson:
@nanetteCodes is it possible to make this part of `keyBindingFn`? thats the first choice before falling back to `handleKeyDown`. also you need to account for arrow up

nanetteCodes:
@efjacobson please review again

src/plugins/blocksmith/components/blocksmith/index.jsx
```diff
@@ -998,6 +1000,9 @@ class Blocksmith extends React.Component {
     const currentBlock = getBlockForKey(contentState, anchorKey);
     const previousBlock = contentState.getBlockBefore(anchorKey);
     const nextBlock = contentState.getBlockAfter(anchorKey);
+    const firstLine = isOnFirstLineOfBlock(editorState, 'line-length-tester');
+    const lastLine = isOnLastLineOfBlock(editorState, 'line-length-tester');
```
efjacobson: @nanetteCodes as you've seen, the process for determining this is pretty intensive and should not be done unless necessary - so defer these until we know for sure we need it

src/plugins/blocksmith/components/blocksmith/index.jsx
```diff
@@ -1041,6 +1046,16 @@ class Blocksmith extends React.Component {
           }
         }
         break;
+      case 'ArrowDown':
+        if (!nextBlock && lastLine) {
```
efjacobson: @nanetteCodes the deferred calculation would be `if (!nextBlock && isOnLastLineOfBlock(editorState, FOCUS_PLUGIN_TEST_CLASS)) {`

src/plugins/blocksmith/plugins/focus/index.js
```diff
@@ -196,14 +151,11 @@ export default (config = {}) => {
...
       const editorState = getEditorState();
-      const lines = getLines(editorState, testClass);
-      let offsetAtStartOfLastLine = 0;
-      for (let i = 0; i < lines.length - 1; i += 1) {
-        offsetAtStartOfLastLine += lines[i].length;
-      }
-      if (lines.length <= 1 || editorState.getSelection().getAnchorOffset() + 1 > offsetAtStartOfLastLine) {
+      const lastLine = isOnLastLineOfBlock(editorState, testClass);
+      
+      if (lastLine) {
```
efjacobson: @nanetteCodes just put the function call inside the if. then it reads almost like a normal human sentence "if is on last line of block"

efjacobson:
closing in favor of https://github.com/triniti/cms-js/pull/186, it has this work and the removal of the focus plugin

PR #187, opened 2020-02-05
fluid height for hashtag picker field

nanetteCodes:


src/plugins/taxonomy/components/hashtags-picker-field/index.jsx
```diff
@@ -78,6 +85,7 @@ const HashtagsPickerField = ({
         className="hidden-options-toggle"
         closeOnSelect={false}
         creatable
+        styles={customStyles}
```
efjacobson: @nanetteCodes just put the object straight in here, dont need the customStyles var

PR #186, opened 2020-02-04
Arrow keys block bug and remove focus plugin

efjacobson:
i made some small changes to https://github.com/triniti/cms-js/tree/arrow-keys-block-bug and then removed the focus plugin and here we are. the focus plugin stopped working a while back and no one noticed so clearly its not important

@nanetteCodes can you make sure i didnt mess up any of your stuff?
@arunkarnati can you review the removal of the focus plugin?

PR #190, opened 2020-02-07
Arrow keys block bug and remove focus plugin again

efjacobson:
https://github.com/triniti/cms-js/pull/186 got janked up after the release

notable new commits:
https://github.com/triniti/cms-js/pull/190/commits/73b47477150cbd7a8f0e861d0a321104d33a2e86
https://github.com/triniti/cms-js/pull/190/commits/7371c0e6cb5c21a447f9b5e151e432e7bd7f7de3


src/plugins/blocksmith/utils/getLines.js
```diff
@@ -0,0 +1,107 @@
...
+ * some keyboard navigation behavior.
+ *
+ * @param {EditorState} editorState - a state instance of a DraftJs Editor
+ *
+ * @returns {Array}
+ */
+export default (editorState) => {
+  const anchorKey = editorState.getSelection().getAnchorKey();
+  const offsetKey = DraftOffsetKey.encode(anchorKey, 0, 0);
+  const currentBlockNode = document.querySelectorAll(`[data-offset-key="${offsetKey}"]`)[0];
```
arunkarnati: @efjacobson @nanetteCodes - instead of getting the width of the`div` node, get the width of `span` under it so it is really accurate and works all the time. 

Here's a screengrab of a specific scenario where the lines are not created accurately because of the width differences between the `div` and `span` tag. 



```suggestion
  const currentBlockNode = document.querySelector(`[data-offset-key="${offsetKey}"] span`);
```
I tested and it worked as expected. Please test it before merging the suggestion.

efjacobson:
@arunkarnati please try again with recent commit

src/plugins/blocksmith/utils/getLines.js
```diff
@@ -25,38 +27,34 @@ const getLinesBrokenByLineBreaks = (txt) => {
  * Calculates the number of lines for a given string using an actual dom node and what happens to
  * it as text is appended.
  *
- * @param {number} spanWidth - the actual width of the text block's dom node
- * @param {string} text      - a string to check for wrapped text
+ * @param {number} node - a node to fill with text to find the wrapped lines
```
arunkarnati: @efjacobson - change the param type to `Object`

src/plugins/blocksmith/utils/getLines.js
```diff
@@ -12,11 +12,13 @@ const getLinesBrokenByLineBreaks = (txt) => {
   let text = txt;
   let matchIndex;
   const lines = [];
+
   while (/\r|\n/.test(text)) {
```
arunkarnati: @efjacobson - should we add `\r\n` to this group?

efjacobson: @arunkarnati not sure what you mean

src/plugins/blocksmith/utils/getLines.js
efjacobson: @arunkarnati updated to `HTMLElement`

efjacobson:
@arunkarnati as for the duplication issue, when making rapid changes, draft has some nasty race conditions that can cause that, see

https://github.com/facebook/draft-js/issues/1906
https://github.com/facebook/draft-js/issues/424
and more

i did write some code that reduces the frequency that it happens, but it can still happen `¯\_(ツ)_/¯`

efjacobson:
@arunkarnati i think that commit is already in?

PR #205, opened 2020-02-20
Blocksmith multi select

efjacobson:


src/plugins/blocksmith/components/blocksmith/index.jsx
```diff
@@ -517,15 +530,15 @@ class Blocksmith extends React.Component {
...
           },
         };
       case 'unstyled':
         return {
-          component: DraggableTextBlock,
+          component: TextBlockWrapper,
           contentEditable: !readOnly,
           props: {
-            getReadOnly: this.getReadOnly,
+            ...blockConfig,
```
efjacobson: @jalopez13 i think you can simplify all the prop drilling by just loading the component with its props and sending it down `blockButtonsComponent: (props) => <BlockButtons {...props} {...blockConfig} />`

and in the placeholder do

```
const BlockButtons = blockButtonsComponent;

return (
...
{!isHidden && <BlockButtons />}
...
)
```

PR #209, opened 2020-03-13
Blocksmith refactor multi select

jalopez13:
Added multi-select drag n drop to blocksmith blocks.


efjacobson:
@jalopez13 change to go into release-next and lmk when you have addressed the issues we discussed yesterday. namely:
- color when another color already exists (update/unpublished block)
- always paste in the order the blocks are already in (not in the order they were clicked)
- should not have to click into editor to see buttons (with no regression of https://github.com/wb-apps/tmz-cms/issues/982)
- block content (text etc) should not be obstructed by the buttons
- any "selected for drag" state or styling should be wiped out on any change

PR #222, opened 2020-05-02
922 cursor

rsumilang:


efjacobson:
@rsumilang seems like a weird branch situation here...quite a few unrelated file changes

src/plugins/blocksmith/utils/index.js
```diff
@@ -33,6 +33,7 @@ import isFirstListBlockFn from './isFirstListBlock';
 import isLastListBlockFn from './isLastListBlock';
 import loadFacebookSDKFn from './loadFacebookSDK';
 import loadTwitterSDKFn from './loadTwitterSDK';
+import moveSelectionToEndOfBlockFn from './moveSelectionToEndOfBlock';
```
efjacobson, line 36: @rsumilang there is a util for this already, `selectBlock`, use it with `selectionTypes.END` to do the same as `moveSelectionToEndOfBlock`

arunkarnati:
@rsumilang - to fix the weird changeset issue you can either rebase commits onto master or just check out a new branch to apply the changes (since the changeset is not that bad)

Also, I don't think there is any need for a `release-next` branch in a package as we don't batch the changes for one big release.

rsumilang:
Moved changes to `922-cursor-new`.

PR #223, opened 2020-05-04
Add custom code tab to timeline screen.

albertgcsula:
Add custom code tab to timeline screen, if schema has the mixin.

### DIFF
--- a/src/plugins/curator/screens/timeline/index.js
+++ b/src/plugins/curator/screens/timeline/index.js
@@ -14,6 +14,7 @@ class TimelineScreen extends AbstractNodeScreen {
   getTabs() {
     return [
       'details',
+      delegate.getSchemas().node.hasMixin('triniti:common:mixin:custom-code') && 'code',
       'taxonomy',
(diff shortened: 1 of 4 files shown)


src/plugins/curator/screens/timeline/index.js
```diff
@@ -14,6 +14,7 @@ class TimelineScreen extends AbstractNodeScreen {
   getTabs() {
     return [
       'details',
+      delegate.getSchemas().node.hasMixin('triniti:common:mixin:custom-code') && 'code',
```
gdbrown, line 17: @albertgcsula use schemas that is already available (like the seo check below)

albertgcsula: @gdbrown updated.

efjacobson:
wait this needs something else

albertgcsula:
> @albertgcsula you need to add the `code` bit to the curator `routes.js`

Thanks for catching that @efjacobson The code is updated.

PR #228, opened 2020-05-16
History revert

rsumilang:


src/plugins/pbjx/components/revert-view-modal/RevertDetails.jsx
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * This will keep the whole object or list for a key. The reason is trying
+ * to revert specific keys in a block list or multiselect list is difficult
+ * because the data might have changed greatly. So we just put it back
+ * completely to the original state.
+ * 
+ * @param {Object} newNode 
+ * @param {Object} origNode 
+ */
+const fullMapsAndLists = (newNode, origNode) => {
```
efjacobson, line 18: @rsumilang getting an eslint warning on this guy

src/plugins/pbjx/components/revert-view-modal/RevertDetails.jsx
```diff
@@ -0,0 +1,65 @@
...
+      newNode[key] = origNode[key];
+    }
+  }
+  return newNode;
+}
+
+const RevertDetails = ({ event, isFieldSelected, onSelectField: handleSelectField }) => {
+  const schema = event.schema();
+
+  if (schema.hasMixin('gdbots:ncr:mixin:node-updated')) {
```
efjacobson, line 30: @rsumilang is it possible for this to be anything else? the event stream component only shows the button when it is an updated event

src/plugins/pbjx/components/revert-view-modal/index.jsx
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    this.isFieldSelected = this.isFieldSelected.bind(this);
+    this.handleSelectField = this.handleSelectField.bind(this);
+    this.handleRevertSwal = this.handleRevertSwal.bind(this);
+  }
+
+  isFieldSelected(id) {
+    const { selected } = this.state;
+    for (const { id: itemId } of selected) {
```
efjacobson, line 54: @rsumilang similar eslint warning here

src/plugins/pbjx/components/history/delegate.js
```diff
@@ -28,6 +34,24 @@ export default (dispatch) => ({
...
+    selected.forEach((item) => {
+      const { id, value } = item;
+      if (id === 'blocks') {
+        const canvasBlocks = value.filter((x) => x !== null).map((x) => Message.fromObject(x));
+        const editorState = convertToEditorState(canvasBlocks);
+        dispatch(storeEditor(formName, editorState, true));
+      } else {
+        const fieldType = node.schema().getField(id).getType().typeName.name;
+        if (fieldType === 'DATE_TIME' && value) {
+          dispatch(change(formName, camelCase(id), new Date(value)));
```
efjacobson, line 47: @rsumilang this doesnt work for live_m3u8_url because that camelCases to liveM3U8Url

src/plugins/curator/screens/teaser/Form.jsx
```diff
@@ -50,14 +50,14 @@ const getFieldsComponent = (type) => {
   }
 };
 
-const Form = ({ node, tab, type, isEditMode }) => {
+const Form = ({ form, node, tab, type, isEditMode }) => {
   const [TeaserFields, setTeaserFields] = useState(null);
   const streamId = StreamId.fromString(`${node.schema().getCurie().getMessage()}.history:${node.get('_id')}`);
```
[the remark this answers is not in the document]
rsumilang: I don't know what you mean by creating it's own stream. Example?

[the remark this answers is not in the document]
efjacobson: @rsumilang right now all these Form components are making their own streamIds: https://github.com/triniti/cms-js/blob/5bcd11804a7f9561b5a05946505beae199dfc7b2/src/plugins/iam/screens/user/Form.jsx#L16

but if we are going to pass the node into the History component then the History component can create the streamId inside itself like so

```
const streamId = StreamId.fromString(`${node.schema().getCurie().getMessage()}.history:${node.get('_id')}`);
```

and then the Forms dont have to worry about it, only the History component does it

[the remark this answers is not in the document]
efjacobson: like it could happen inside the delegate because that delegate will have access to the node via ownProps https://github.com/triniti/cms-js/blob/master/src/plugins/pbjx/components/history/delegate.js#L14

src/plugins/pbjx/components/revert-view-modal/index.jsx
rsumilang: i'm not getting an eslint issue here?

src/plugins/blocksmith/actions/storeEditor.js
```diff
@@ -1,7 +1,8 @@
 import { actionTypes } from '../constants';
 
-export default (formName, editorState) => ({
+export default (formName, editorState, isDirty) => ({
```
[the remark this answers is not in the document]
rsumilang: Ah it wasn't clear to me thats the way to do it. Looked like storeEditor was just missing one of the blocksmithState props so I added it. I can change.

src/plugins/pbjx/components/revert-view-modal/RevertDetails.jsx
rsumilang: not getting it on mine

src/plugins/blocksmith/reducers/index.js
```diff
@@ -1,4 +1,5 @@
 import createReducer from '@triniti/app/createReducer';
+import isBoolean from 'lodash/isBoolean';
```
efjacobson, line 2: @rsumilang not used

src/plugins/pbjx/components/history/delegate.js
```diff
@@ -1,33 +1,68 @@
...
+      const idFormatted = id === 'live_m3u8_url' ? 'liveM3u8Url' : camelCase(id);
+
+      if (id === 'blocks') {
+        const canvasBlocks = value.filter((x) => x !== null).map((x) => Message.fromObject(x));
+        const editorState = convertToEditorState(canvasBlocks);
+        dispatch(storeEditor(formName, editorState));
+        dispatch(dirtyEditor(formName));
+      } else {
+        const fieldType = node.schema().getField(id).getType();
+        if (fieldType instanceof DateTimeType && value) {
```
efjacobson, line 57: @rsumilang this doesnt work if the date was removed

src/plugins/pbjx/components/history/index.jsx
```diff
@@ -3,8 +3,6 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
```
efjacobson, line 3: @rsumilang several eslint warnings about props validation in this component

src/plugins/pbjx/components/history/index.jsx
```diff
@@ -22,18 +21,19 @@ class History extends React.Component {
       response: PropTypes.object,
       status: PropTypes.string,
     }).isRequired,
+    isEditMode: PropTypes.bool.isRequired,
     getUser: PropTypes.func.isRequired,
-    schema: PropTypes.instanceOf(Schema).isRequired,
-    streamId: PropTypes.instanceOf(StreamId).isRequired,
+    formName: PropTypes.string.isRequired,
```
efjacobson, line 26: @rsumilang alpha sort props please

src/plugins/pbjx/components/history/selector.js
```diff
@@ -1,4 +1,5 @@
 import NodeRef from '@gdbots/schemas/gdbots/ncr/NodeRef';
+import isGranted from '@triniti/cms/plugins/iam/selectors/isGranted';
```
efjacobson, line 2: @rsumilang not used

src/plugins/pbjx/components/revert-modal/index.jsx
```diff
@@ -0,0 +1,114 @@
+import PropTypes from 'prop-types';
```
efjacobson, line 1: @rsumilang got some eslint jazz going on in this component

src/plugins/pbjx/components/history/index.jsx
rsumilang: Screen Shot 2020-06-09 at 2 26 55 PM

I don't see any.

src/plugins/pbjx/components/revert-modal/index.jsx
rsumilang: Screen Shot 2020-06-09 at 2 26 55 PM

I don't see any.

src/plugins/pbjx/components/history/index.jsx
efjacobson: Screen Shot 2020-06-09 at 3 44 33 PM

src/plugins/pbjx/components/revert-modal/index.jsx
efjacobson: Screen Shot 2020-06-09 at 3 45 11 PM

src/plugins/pbjx/components/history/delegate.js
rsumilang: actually it does because the `} else {` block code catches it.

src/plugins/pbjx/utils/filterData.js
```diff
@@ -29,6 +29,12 @@ const omitKeys = [
   'word_count',
 ];
 
+const omitRevertKeys = omitKeys.concat([
+  'jwplayer_synced_at',
```
efjacobson, line 33: @rsumilang think we need kaltura_synced_at here too

src/plugins/pbjx/components/history/index.jsx
```diff
@@ -47,21 +50,26 @@ class History extends React.Component {
...
-      schema,
+      formName,
+      node,
     } = this.props;
 
     return (
       <EventStream
         events={events}
         getUser={getUser}
+        formName={formName}
```
efjacobson, line 64: @rsumilang doesnt look like this prop is used by EventStream

src/plugins/pbjx/components/history/index.jsx
```diff
@@ -47,21 +50,26 @@ class History extends React.Component {
...
+        node={node}
```
efjacobson, line 65: @rsumilang nor this one

src/plugins/pbjx/components/history/delegate.js
```diff
@@ -1,33 +1,68 @@
...
       since,
     }));
   },
 
+  /**
+   * Handles Revert.
+   *
+   * @param {array} selected
+   */
+  handleRevert: (selected) => {
```
efjacobson, line 44: @rsumilang having this pbjx component directly import and use blocksmith stuff is making me itchy. can you turn lines 51-54 into a blocksmith util `revertBlocks` (or `restoreBlocks` depending on where we land on that) that is then passed into the component. pbjx is low level, it would be better if it were completely unaware of blocksmith

src/plugins/pbjx/components/history/delegate.js
```diff
@@ -1,33 +1,68 @@
...
+    const { node, formName } = ownProps;
+    selected.forEach((item) => {
+      const { id, value } = item;
+      const idFormatted = id === 'live_m3u8_url' ? 'liveM3u8Url' : camelCase(id);
+
+      if (id === 'blocks') {
+        const canvasBlocks = value.filter((x) => x !== null).map((x) => Message.fromObject(x));
```
efjacobson, line 51: @rsumilang can prevent looping this twice with reduce if you're feeling fancy

src/plugins/pbjx/components/history/index.jsx
rsumilang: I removed formName and node but it's a little weird because I then have to remove them from the required props. However they are required because the delegate needs them but now it's not listed anywhere. Hmm.

src/plugins/pbjx/components/history/delegate.js
rsumilang: If I create `revertBlocks` as a blocksmith util, then the history component will need to import it and pass it into its props. Doesn't feel like this solution gets to rid of knowing about blocksmith completely and introduces that other weird thing of passing in a prop to a component that is now required but cannot be defined as required on the component that accepts it because it doesn't use it, the delegate does.

rsumilang: Perhaps it would make more sense if the History component was its own plugin. But idk if I care that much.

efjacobson: @rsumilang it would be imported by the forms and passed in. that way this pbjx component doesnt know about a blocksmith util, just a blocks util

the delegate/component required props is what it is, not that big of a deal we just do the best we can

src/plugins/pbjx/components/history/index.jsx
efjacobson: @rsumilang its a wrinkle of required props, not that big of a deal

rsumilang:
I think it's just semantics between "revert" and "restore". Probably the right message would be "nothing to revert" if you tried to revert to something that was the current state or maybe not even show the button if the states match. However I feel like it's currently fine for the intended use.

src/plugins/pbjx/components/history/delegate.js
rsumilang: Scratch that comment about the history component importing it in. It would be every screens form and passing to the history component. However the delegate still requires thus making it have some knowledge of it unless it was some abstracted handler. I mean, yes pbjx feels like a lower level thing but it has knowledge of form controls and blocksmith feels like a form thing.

efjacobson:
@rsumilang i spoke with Sam and gave him a demo, we came up with the following items

1. we do want to do the inverse of what this is doing now. as in, this should be a restore button instead of a revert button. its much more clear that way, the user will just look at the field values in the modal and know that they will change to those values, instead of having to be aware of what they were _before_ and know that that will become the actual value. it will be "what you see is what you get"

2. the button should not appear for the most recent event in the stream, because there will be nothing to change

3. lets hide any values from the modal that are not different than what is currently set on the node. having them checked and clickling apply should not do anything in that scenario anyways so its confusing. please also add a disclaimer after the Revert Fields header, something like: `Revert Fields (fields not shown are current with the article/video/blah value)`

4. this one may not be relevant because of the other changes, but there is a bit of a bug here where you can revert a field to what it already is. in this example (http://recordit.co/zDd0CiDllH) you can see the article starts out with no classification, then the classification is reverted (to null), then the save button is enabled and saving causes an updated event with no changes to happen

rsumilang:
I don't understand what you mean by making it do the inverse? It seems like it currently is what you see is what you get because the modal shows you what it will revert to. Can you explain?

efjacobson:
@rsumilang sure, ill give it a shot, otherwise we might want to hop on a call.

in this video (http://recordit.co/FMndXo2fqa) i
1. make a new article
2. update the title and classification
3. update the title
4. go to revert the classification. what i see here is classification:hero. what i get is classification:null

if it was a restore button, what you see in the modal can stay the same. but instead of changing the value to null, it would change it to hero

rsumilang:
Ah, your video points out that there is an issue setting the drop down box to the selected item. I tested just now and it appears to work fine with clearing the dropdown box but not selecting the right item. I'll fix that. Otherwise it's currently doing what it's supposed to which is setting the values you have checked to their appropriates fields regardless of what they are (wysiwyg). That being said, I'm not convinced I need to rename everything to Restore.

efjacobson:
@rsumilang can you connect with Sam on that? i need to focus on some other stuff. the notes i put into https://github.com/triniti/cms-js/pull/228#issuecomment-644883799 are what Sam and i came up with this morning

rsumilang:
@efjacobson Spoke to Sam, the Revert/Restore naming doesn't matter so I'll keep it as. I think there is a misunderstanding of how the revert works due the dropdown issue.

src/plugins/pbjx/components/event-stream/index.jsx
```diff
@@ -81,14 +87,19 @@ class EventStream extends React.Component {
...
       getUser,
       response,
       status,
       exception,
+      onRevert: handleRevert,
     } = this.props;
 
     const { allEvents, loadMore } = this.state;
 
+    let index = 0;
```
efjacobson, line 101: @rsumilang `map` has access to the array index https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

rsumilang:
Yes I have some changes in local. Ticket is currently in WIP and should be updated today.

rsumilang:
@efjacobson ready

src/plugins/pbjx/components/history/delegate.js
```diff
@@ -5,6 +5,60 @@ import camelCase from 'lodash/camelCase';
...
+      formValue = [];
+      /* eslint-disable no-restricted-syntax */
+      for (const [key, keyValue] of Object.entries(value)) {
+        formValue.push({
+          key,
+          value: keyValue,
+        });
+      }
+    } else if (id === 'hashtags' && Array.isArray(value)) {
+      formValue = value.reduce((accumulator, currentValue) => {
```
efjacobson, line 45: @rsumilang this can just be a map eh?

src/plugins/apollo/screens/poll/Form.jsx
```diff
@@ -1,23 +1,29 @@
...
   switch (tab) {
     case 'history':
-      return <History schema={schemas.getNodeHistoryRequest} streamId={streamId} />;
+      return (
+        <History
+          isEditMode={isEditMode}
+          formName={form}
+          node={node}
+          schema={schemas.getNodeHistoryRequest}
+          setBlocks={setBlocks}
```
efjacobson, line 24: @rsumilang polls dont have blocks so i dont think you need to provide this. can just default to noop

src/plugins/blocksmith/utils/setBlocks.js
```diff
@@ -0,0 +1,19 @@
+import dirtyEditor from '@triniti/cms/plugins/blocksmith/actions/dirtyEditor';
+import storeEditor from '@triniti/cms/plugins/blocksmith/actions/storeEditor';
+import convertToEditorState from '@triniti/cms/plugins/blocksmith/utils/convertToEditorState';
+import Message from '@gdbots/pbj/Message';
+
+export default (dispatch, formName, blocks, isDirty = false) => {
```
efjacobson, line 6: @rsumilang why is there isDirty? looks like it is only being used like `setBlocks(dispatch, formName, formValue, true);`

src/plugins/boost/screens/sponsor/Form.jsx
```diff
@@ -7,15 +7,22 @@ import History from '@triniti/cms/plugins/pbjx/components/history';
...
   switch (tab) {
     case 'history':
-      return <History schema={schemas.getNodeHistoryRequest} streamId={streamId} />;
+      return (
+        <History
+          isEditMode={isEditMode}
+          formName={form}
+          node={sponsor}
+          schema={schemas.getNodeHistoryRequest}
+          setBlocks={setBlocks}
```
efjacobson, line 23: @rsumilang same as poll. i think only people, pages, and articles can have blocks

src/plugins/blocksmith/utils/setBlocks.js
rsumilang: I wrote it like this because there is somewhere in the block editor that is also setting blocks but not dirtying it when loading the contents. It can possibly use this.

rsumilang:
@efjacobson is this good?

efjacobson:
@rsumilang yeah looks good, can you just confirm with @SamWhiteTMZ about showing fields that dont actually have any changes? like, as it is now you can "revert" something to the same title it already has, and the modal says you successfully reverted even though the form didnt change

efjacobson:
> @rsumilang yeah looks good, can you just confirm with @SamWhiteTMZ about showing fields that dont actually have any changes? like, as it is now you can "revert" something to the same title it already has, and the modal says you successfully reverted even though the form didnt change

@SamWhiteTMZ @rsumilang can we make a final decision on this so this can get wrapped up?

rsumilang:
@efjacobson a decision was made yesterday to hide changes that are currently the same. currently wip.

rsumilang:
@efjacobson Added the following things:

1. Disabled the button if all the values in the form modal are the same as the node.
2. Hid the values that match the same as the node in the modal.

Changes are ready for review.

src/plugins/pbjx/components/history/delegate.js
```diff
@@ -51,8 +51,12 @@ const toFormValue = (id, value) => {
 
   let formValue = value;
 
-  if ((id.endsWith('refs') || ['channel_ref', 'sponsor_ref'].includes(id)) && !Array.isArray(value)) {
-    formValue = [NodeRef.fromString(value)];
+  if (id.endsWith('refs') || ['channel_ref', 'sponsor_ref'].includes(id)) {
```
efjacobson, line 54: @rsumilang had a chat with Greg last night about simplifying the cms in several regards including forms, and we agree the logic in this delegate has gotten way too hairy. its a gnarly situation but we gotta do our part to keep it from getting any worse. i think we have to do the transformer strategy that we discussed in slack to minimize all these edge cases

rsumilang: Are you talking about looking up the node to figure out the field types? If not, can you provide me more details.

efjacobson: @rsumilang sure, its like we discussed before: there should be a transformer util for each field that can transform a field value into the form value and the form value into the field value. then these utils would be used by both the subscribers and this history jazz

rsumilang:
@efjacobson this is ready for you

src/plugins/pbjx/reducers/index.js
```diff
@@ -39,7 +39,19 @@ const onResponseCleared = (prevState = {}, action) => {
   return state;
 };
 
+const onUpdateResponseNode = (prevState = {}, { node, channel, curie }) => {
+  const state = { ...prevState };
+  const curieStr = `${curie}`;
+  const newResponse = state[channel][curieStr].response.clone();
+
+  newResponse.set('node', node);
```
efjacobson, line 47: @rsumilang this is pretty weird man. now we are hijacking the pbjx state to force the form to change? doesnt feel right

rsumilang: @efjacobson yes, now it's not as sensitive to forms changing and having to keep an eye on new form components. changing the pbjx state allows for each subscriber to do its thing.

src/plugins/curator/screens/gallery/Form.jsx
```diff
@@ -33,7 +32,16 @@ const Form = ({ node: gallery, form, isEditMode, tab }) => {
...
     case 'history':
-      return <History schema={schemas.getNodeHistoryRequest} streamId={streamId} />;
+      return (
+        <History
+          isEditMode={isEditMode}
+          formName={form}
+          node={gallery}
+          nodeRequest={getNodeRequestState.request}
+          schema={schemas.getNodeHistoryRequest}
+          setBlocks={setBlocks}
```
efjacobson, line 42: @rsumilang galleries cant have blocks

src/plugins/curator/screens/teaser/Form.jsx
```diff
@@ -50,14 +50,21 @@ const getFieldsComponent = (type) => {
...
     case 'history':
-      return <History schema={schemas.getNodeHistoryRequest} streamId={streamId} />;
+      return (
+        <History
+          isEditMode={isEditMode}
+          formName={form}
+          node={node}
+          nodeRequest={getNodeRequestState.request}
+          schema={schemas.getNodeHistoryRequest}
+          setBlocks={setBlocks}
```
efjacobson, line 65: @rsumilang teasers dont have blocks

src/plugins/curator/screens/timeline/Form.jsx
```diff
@@ -30,7 +29,16 @@ const Form = ({ form, node, tab, isEditMode }) => {
...
     case 'history':
-      return <History schema={schemas.getNodeHistoryRequest} streamId={streamId} />;
+      return (
+        <History
+          isEditMode={isEditMode}
+          formName={form}
+          node={node}
+          nodeRequest={getNodeRequestState.request}
+          schema={schemas.getNodeHistoryRequest}
+          setBlocks={setBlocks}
```
efjacobson, line 39: @rsumilang timelines dont have blocks

src/plugins/curator/screens/widget/Form.jsx
```diff
@@ -75,15 +75,23 @@ const getFieldsComponent = (type) => {
...
     case 'history':
-      return <History schema={schemas.getNodeHistoryRequest} streamId={streamId} />;
+      return (
+        <History
+          isEditMode={isEditMode}
+          formName={form}
+          node={node}
+          nodeRequest={getNodeRequestState.request}
+          schema={schemas.getNodeHistoryRequest}
+          setBlocks={setBlocks}
```
efjacobson, line 92: @rsumilang dont need setBlocks here

src/plugins/dam/screens/asset/Form.jsx
```diff
@@ -66,7 +65,16 @@ const Form = ({ node: asset, tab, isEditMode, type }) => {
...
     case 'history':
-      return <History schema={getNodeHistoryRequest} streamId={streamId} />;
+      return (
+        <History
+          isEditMode={isEditMode}
+          formName={form}
+          node={asset}
+          nodeRequest={getNodeRequestState.request}
+          schema={getNodeHistoryRequest}
+          setBlocks={setBlocks}
```
efjacobson, line 75: @rsumilang

src/plugins/iam/screens/app/Form.jsx
```diff
@@ -53,7 +52,16 @@ const Form = ({ node, tab, type, isEditMode }) => {
...
     case 'history':
-      return <History schema={schemas.getNodeHistoryRequest} streamId={streamId} />;
+      return (
+        <History
+          isEditMode={isEditMode}
+          formName={form}
+          node={node}
+          nodeRequest={getNodeRequestState.request}
+          schema={schemas.getNodeHistoryRequest}
+          setBlocks={setBlocks}
```
efjacobson, line 62: @rsumilang

src/plugins/iam/screens/role/Form.jsx
```diff
@@ -7,15 +7,23 @@ import Message from '@gdbots/pbj/Message';
...
     case 'history':
-      return <History schema={schemas.getNodeHistory} streamId={streamId} />;
+      return (
+        <History
+          isEditMode={isEditMode}
+          formName={form}
+          node={role}
+          nodeRequest={getNodeRequestState.request}
+          schema={schemas.getNodeHistory}
+          setBlocks={setBlocks}
```
efjacobson, line 24: @rsumilang

src/plugins/iam/screens/user/Form.jsx
```diff
@@ -23,7 +22,16 @@ const Form = ({ node: user, tab, isEditMode }) => {
...
     case 'history':
-      return <History schema={schemas.getNodeHistory} streamId={streamId} />;
+      return (
+        <History
+          isEditMode={isEditMode}
+          formName={form}
+          node={user}
+          nodeRequest={getNodeRequestState.request}
+          schema={schemas.getNodeHistory}
+          setBlocks={setBlocks}
```
efjacobson, line 32: @rsumilang

src/plugins/ncr/screens/node/AbstractDelegate.js
```diff
@@ -299,7 +299,8 @@ export default class AbstractDelegate {
     const node = this.component.props.getNodeRequestState.response.get('node');
     const formEvent = this.createFormEvent({}, { node, form: this.getFormName() });
     this.pbjx.trigger(formEvent.getMessage(), SUFFIX_INIT_FORM, formEvent);
-    return formEvent.getData();
+    // Force dirty is used for history revert.
+    return { ...formEvent.getData(), _forceDirty: false };
```
efjacobson, line 303: @rsumilang i dont think is is necessary right, since dispatching it with the change will make it not pristine

src/plugins/news/screens/article/index.jsx
```diff
@@ -26,8 +26,8 @@ class ArticleScreen extends AbstractNodeScreen {
   };
 
   getFormRenderProps() {
-    const { blocksmithState } = this.props;
-    return { blocksmithState };
+    const { blocksmithState, getNodeRequestState } = this.props;
+    return { blocksmithState, getNodeRequestState };
```
efjacobson, line 30: @rsumilang i think you need this in a lot of other places? yes - all the non article history tabs are crashing

src/plugins/notify/screens/notification/Form.jsx
```diff
@@ -51,16 +51,22 @@ const getFieldsComponent = (type) => {
...
     case 'history':
-      return <History schema={schemas.getNodeHistoryRequest} streamId={streamId} />;
+      return (
+        <History
+          isEditMode={isEditMode}
+          formName={form}
+          node={node}
+          nodeRequest={getNodeRequestState.request}
+          schema={schemas.getNodeHistoryRequest}
+          setBlocks={setBlocks}
```
efjacobson, line 67: @rsumilang

src/plugins/ovp/screens/video/Form.jsx
```diff
@@ -12,18 +12,26 @@ import PropTypes from 'prop-types';
...
     case 'history':
-      return <History schema={schemas.getNodeHistoryRequest} streamId={streamId} />;
+      return (
+        <History
+          isEditMode={isEditMode}
+          formName={form}
+          node={video}
+          nodeRequest={getNodeRequestState.request}
+          schema={schemas.getNodeHistoryRequest}
+          setBlocks={setBlocks}
```
efjacobson, line 32: @rsumilang

src/plugins/pbjx/components/history/delegate.js
```diff
@@ -1,37 +1,138 @@
...
+          updatedNode.addToMap(id, k, v);
+        }
+      }
+    });
+
+    this.dispatch(updateReponseNode(updatedNode, nodeRequest.schema().getCurie()));
+
+    setTimeout(() => {
+      this.dispatch(change(formName, '_forceDirty', true));
+    }, 1000);
```
efjacobson, line 95: @rsumilang dont think a whole second is necessary here, probably fine to just drop it at the bottom of the stack

src/plugins/sys/screens/flagset/Form.jsx
```diff
@@ -2,18 +2,26 @@ import React from 'react';
...
     case 'history':
-      return <History schema={schemas.getNodeHistory} streamId={streamId} />;
+      return (
+        <History
+          isEditMode={isEditMode}
+          formName={form}
+          node={flagset}
+          nodeRequest={getNodeRequestState.request}
+          schema={schemas.getNodeHistory}
+          setBlocks={setBlocks}
```
efjacobson, line 22: @rsumilang

src/plugins/sys/screens/picklist/Form.jsx
```diff
@@ -2,19 +2,27 @@ import React from 'react';
...
     case 'history':
-      return <History schema={schemas.getNodeHistory} streamId={streamId} />;
+      return (
+        <History
+          isEditMode={isEditMode}
+          formName={form}
+          node={picklist}
+          nodeRequest={getNodeRequestState.request}
+          schema={schemas.getNodeHistory}
+          setBlocks={setBlocks}
```
efjacobson, line 23: @rsumilang

src/plugins/sys/screens/redirect/Form.jsx
```diff
@@ -23,7 +22,16 @@ const Form = ({
...
     case 'history':
-      return <History schema={schemas.getNodeHistoryRequest} streamId={streamId} />;
+      return (
+        <History
+          isEditMode={isEditMode}
+          formName={form}
+          node={redirect}
+          nodeRequest={getNodeRequestState.request}
+          schema={schemas.getNodeHistoryRequest}
+          setBlocks={setBlocks}
```
efjacobson, line 32: @rsumilang

src/plugins/taxonomy/screens/category/Form.jsx
```diff
@@ -10,21 +10,29 @@ import History from '@triniti/cms/plugins/pbjx/components/history';
...
     case 'history':
-      return <History schema={schemas.getNodeHistoryRequest} streamId={streamId} />;
+      return (
+        <History
+          isEditMode={isEditMode}
+          formName={form}
+          node={category}
+          nodeRequest={getNodeRequestState.request}
+          schema={schemas.getNodeHistoryRequest}
+          setBlocks={setBlocks}
```
efjacobson, line 33: @rsumilang

src/plugins/taxonomy/screens/channel/Form.jsx
```diff
@@ -10,21 +10,29 @@ import History from '@triniti/cms/plugins/pbjx/components/history';
...
     case 'history':
-      return <History schema={schemas.getNodeHistoryRequest} streamId={streamId} />;
+      return (
+        <History
+          isEditMode={isEditMode}
+          formName={form}
+          node={channel}
+          nodeRequest={getNodeRequestState.request}
+          schema={schemas.getNodeHistoryRequest}
+          setBlocks={setBlocks}
```
efjacobson, line 33: @rsumilang

src/plugins/ncr/screens/node/AbstractDelegate.js
rsumilang: No it needs it because dispatching a change to a field that doesn't exist won't do anything.

src/plugins/pbjx/components/history/delegate.js
rsumilang: @efjacobson I can maybe change it to half a second but it needed something. Firing it immediately after did not alter the form.

src/plugins/ncr/screens/node/AbstractDelegate.js
efjacobson: @rsumilang odd, it works fine for me when i just `return formEvent.getData();`

rsumilang: @efjacobson hmm indeed... seems to work now after another sanity check

src/plugins/pbjx/components/history/delegate.js
efjacobson: @rsumilang yeah redux form is like that, it needs to process the changes from changing the request node first. i expect 

```
setTimeout(() => this.dispatch(change(formName, '_forceDirty', true)));
```

will be fine

rsumilang:
@efjacobson anything else to do here?

rsumilang:
@efjacobson Added the getNodeRequestState() to the AbstractNodeScreen. I went through each screen and it seemed to be working fine. The teaser is fixed too.

src/plugins/pbjx/utils/filterRemoved.js
```diff
@@ -0,0 +1,27 @@
+import isPlainObject from 'lodash/isPlainObject';
+
+/**
+ * Removes "[removed]" entries from nodes changeset.
+ */
+const filterRemoved = (data) => {
+  if (!isPlainObject(data)) {
+    return data;
+  }
+  return Object.keys(data).reduce((accumulator, index) => {
```
efjacobson, line 10: @rsumilang nit: i'd probably change `index` here to `key` since that's what it is

src/plugins/pbjx/utils/fullMapsAndLists.js
```diff
@@ -0,0 +1,21 @@
...
+ * to revert specific keys in a block list or multiselect list is difficult
+ * because the data might have changed greatly. So we just put it back
+ * completely to the original state.
+ *
+ * @param {Object} newNode
+ * @param {Object} origNode
+ */
+export default (newNode, origNode) => {
+  const newerNode = { ...newNode };
+  Object.entries(newerNode).forEach((item) => {
```
efjacobson, line 14: @rsumilang nit: this could be more concise by using `reduce`

src/plugins/pbjx/components/revert-modal/RevertDetails.jsx
```diff
@@ -0,0 +1,42 @@
...
+import PropTypes from 'prop-types';
+import Message from '@gdbots/pbj/Message';
+import RevertPropertiesTable from '../revert-properties-table';
+import { filterRevertableData } from '../../utils/filterData';
+import findNodeDiff from '../../utils/findNodeDiff';
+import fullMapsAndLists from '../../utils/fullMapsAndLists';
+
+const RevertDetails = ({ event, isFieldSelected, isDbValueSameAsNodeValue, onSelectField: handleSelectField }) => {
+  // find properties in node that were removed
+  const newNode = event.get('new_node').toObject();
```
efjacobson, line 11: @rsumilang nit: this bit of code is duplicated a few times, could probably be consolidated into a util

src/plugins/pbjx/components/revert-modal/index.jsx
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+
+  handleRevertSwal() {
+    const { onRevert: handleRevert, onToggleRevertModal: handleToggleRevertModal } = this.props;
+    const { selected } = this.state;
+    handleRevert(selected);
+    RevertModal.revertComplete().then(() => {
+      handleToggleRevertModal();
+    });
```
efjacobson, line 80: @rsumilang nit: fyi you dont have to make a new fn for this, can just do

`RevertModal.revertComplete().then(handleToggleRevertModal)`

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # CHANGELOG
 
 
+## vN.N.N
+* Support for reverting history.
+* Adjusting support to move cursor to the end of last text block.
```
efjacobson, line 6: @rsumilang is this change actually in here?

gdbrown:
@rsumilang can you sync up with the latest in release-next and re-test

rsumilang:
@gdbrown synced up with the latest and every screen seem to load fine.

gdbrown:
@rsumilang can you sync with master / release-next and also, were all of @efjacobson notes addressed?  still shows that changes are requested

rsumilang:
@gdbrown fixed conflict and all notes are addressed.

PR #230, opened 2020-05-30
Fix formatting of jwplayer_media_id

efjacobson:


src/plugins/ovp/services/JwplayerMediaSubscriber.js
```diff
@@ -23,7 +23,9 @@ export default class JwplayerMediaSubscriber extends EventSubscriber {
     data.jwplayerMediaId = node.get('jwplayer_media_id');
     data.jwplayerSyncEnabled = node.get('jwplayer_sync_enabled');
     data.jwplayerSyncedAt = node.has('jwplayer_synced_at')
-      ? moment(node.get('jwplayer_synced_at')).format('MMM DD, YYYY hh:mm A')
+      ? Microtime.fromString(String(`${node.get('jwplayer_synced_at')}000000`))
```
arunkarnati, line 26: @efjacobson - please adjust the util to format a Timestamp Type and use it: https://github.com/triniti/cms-js/blob/master/src/utils/convertReadableTime.js#L3

src/plugins/ovp/services/JwplayerMediaSubscriber.js
```diff
@@ -1,6 +1,6 @@
 import EventSubscriber from '@gdbots/pbjx/EventSubscriber';
```
arunkarnati, line 1: @efjacobson - any reason why we didn't name the subscriber after the mixin name, i.e., `has-media` unlike others?

src/plugins/ovp/services/JwplayerMediaSubscriber.js
arunkarnati: @efjacobson - looks like it does handle the timestamps, so just use it: https://github.com/triniti/cms-js/blob/master/src/utils/convertReadableTime.js#L14

efjacobson:
@arunkarnati good catch on both, thank you. addressed in recent commit https://github.com/triniti/cms-js/pull/230/commits/d20d4ff07cd9c52345d4c6a345df40c0cba0a22f

src/plugins/ovp/services/JwplayerHasMediaSubscriber.js
```diff
@@ -23,7 +24,7 @@ export default class JwplayerMediaSubscriber extends EventSubscriber {
     data.jwplayerMediaId = node.get('jwplayer_media_id');
     data.jwplayerSyncEnabled = node.get('jwplayer_sync_enabled');
     data.jwplayerSyncedAt = node.has('jwplayer_synced_at')
-      ? moment(node.get('jwplayer_synced_at')).format('MMM DD, YYYY hh:mm A')
+      ? convertReadableTime(Microtime.fromString(String(`${node.get('jwplayer_synced_at')}000000`)))
```
arunkarnati, line 27: @efjacobson - I think we just need to give the util a timestamp and it will format the date properly. Please change it to:
```suggestion
      ? 
convertReadableTime(node.get('jwplayer_synced_at'))
```

src/plugins/ovp/services/KalturaEntrySubscriber.js
```diff
@@ -19,12 +20,13 @@ export default class KalturaEntrySubscriber extends EventSubscriber {
   onInitForm(formEvent) {
     const data = formEvent.getData();
     const node = formEvent.getMessage();
-    const syncedAt = node.get('kaltura_synced_at') ? Microtime.fromString(String(`${node.get('kaltura_synced_at')}000000`)) : null;
 
+    data.kalturaSyncedAt = node.get('kaltura_synced_at')
```
arunkarnati, line 24: @efjacobson - same as above:
```suggestion
    data.kalturaSyncedAt = node.has('kaltura_synced_at')
```

src/plugins/ovp/services/KalturaEntrySubscriber.js
```diff
@@ -19,12 +20,13 @@ export default class KalturaEntrySubscriber extends EventSubscriber {
...
+      ? convertReadableTime(Microtime.fromString(String(`${node.get('kaltura_synced_at')}000000`)))
```
arunkarnati, line 25:
```suggestion
      ? convertReadableTime(node.get('kaltura_synced_at'))
```

src/plugins/ovp/services/KalturaEntrySubscriber.js
```diff
@@ -19,12 +20,13 @@ export default class KalturaEntrySubscriber extends EventSubscriber {
...
+      : '';
```
arunkarnati, line 26:
```suggestion
      : null;
```

efjacobson:
@arunkarnati thanks for the feedback, addressed in this diff https://github.com/triniti/cms-js/compare/d20d4ff..b763263

Issue #232, opened 2020-06-24
Copying and pasting content with header tags into blocksmith is creating unsupported block types


arunkarnati:
The custom `blocksmith` component in the application supports only the following block types now:

1. atomic
2. ordered-list-item
3. unordered-list-item
4. unstyled

Convert any other pasted content into blocksmith to `<p>` tags. Check the custom block rendering info on Draft documentation, https://draftjs.org/docs/advanced-topics-custom-block-render-map/

PR #233, opened 2020-06-24
add custom blockRenderMap to avoid creation of header tags when pasting content

arunkarnati:
#232


arunkarnati:
@efjacobson - please retest when you get a chance, thanks!

PR #241, opened 2020-08-12
use correct video asset urls

efjacobson:


efjacobson:
@arunkarnati the issue is that damUrl has versions and qualities that dont really apply to video artifacts, and that stuff is baked into assetid.toFilePath https://github.com/triniti/schemas/blob/master/build/js/src/triniti/dam/AssetId.js#L130

the quality levels are baked into the m3u8 which is the only entrypoint to the underlying hls files, which are themselves the quality levels

also the artifacts have differing extensions so if you wanted to get, say, the thumbnail sprite for a video asset, damUrl would have to know to replace whatever extension with jpg.

Greg and i mulled it over and decided a separate artifact url provider was the way to go https://github.com/triniti/ovp-php/blob/master/src/Ovp/ArtifactUrlProvider.php#L9

PR #242, opened 2020-08-18
Article labels

rsumilang:


src/plugins/ncr/components/labels-picker/index.jsx
```diff
@@ -0,0 +1,225 @@
...
+              <PicklistPicker
+                autoload={false}
+                cache={false}
+                creatable
+                disabled={disabled}
+                isEditMode
+                label="Labels"
+                multi
+                onChange={this.handleChange}
+                picklistId="article-labels"
```
efjacobson, line 186: @rsumilang this should derive 'article' or whatnot from the node

src/plugins/ncr/sagas/updateNodeLabelsFlow.js
```diff
@@ -0,0 +1,37 @@
...
+    successMessage: `Success! The ${startCase(pbj.get('node_ref').getLabel())} was updated.`,
+    verify: (response) => {
+      if (!response.pbj.has('node')) {
+        return false;
+      }
+      const responseLabels = response.pbj.get('node').get('labels', []).sort();
+
+      // check if add_labels was added
+      const addLabels = pbj.get('add_labels', []);
+      if (addLabels.length && !some(addLabels, (label) => includes(responseLabels, label))) {
```
efjacobson, line 24: @rsumilang Array.some and Array.includes are available, dont need lodash

rsumilang:
@efjacobson sam said he was going to check with annie first

src/plugins/ncr/sagas/updateNodeLabelsFlow.js
```diff
@@ -0,0 +1,37 @@
...
+        return false;
+      }
+
+      // check if remove_labels were removed
+      const removeLabels = pbj.get('remove_labels', []);
+      if (removeLabels.length && some(pbj.get('remove_labels', []), (label) => includes(responseLabels, label))) {
+        return false;
+      }
```
hunyoboy, line 32: @rsumilang my questions here is the same as my `some` question above.

rsumilang:
@hunyoboy @efjacobson pr updated

src/plugins/ncr/components/labels-picker/index.jsx
```diff
@@ -183,7 +183,7 @@ class Labels extends React.PureComponent {
                 label="Labels"
                 multi
                 onChange={this.handleChange}
-                picklistId="article-labels"
+                picklistId={`${node.schema().toString().split(':')[4]}-labels`}
```
efjacobson, line 186: @rsumilang `node.schema().getCurie().getMessage()`

src/plugins/ncr/components/labels-picker/index.jsx
```diff
@@ -0,0 +1,225 @@
...
+
+    const values = [];
+    (touched ? selected : node.get('labels', [])).forEach((val) => {
+      values.push({
+        label: val,
+        value: val,
+      });
+    });
+
+    const colors = (label) => labelColors[label] || '#c0c0c0';
```
hunyoboy, line 120: nit: @rsumilang this is a function that returns a single color always? if not then ignore it. what you think if `getColor`?  currently, it sounds like it will return a collection of colors to me.

src/plugins/ncr/components/labels/index.jsx
```diff
@@ -0,0 +1,26 @@
...
+    return {};
+  }
+
+  const color = chroma(labelColors[label]);
+  return {
+    backgroundColor: color.alpha(0.3).css(),
+  };
+};
+
+const Labels = ({ node }) => node.get('labels', []).map((label) => <span className="label" style={createStyle(label)}>{label}</span>);
```
hunyoboy, line 20: @rsumilang nice concise component. need a unique-valued key prop ( key='myKey' ) on each Label to suppress warning on dashboard page.

src/plugins/ncr/components/labels/styles.scss
```diff
@@ -0,0 +1,12 @@
+span.label {
+    margin: .3em;
+    padding: .1em .5em;
+    background-color: white;
+    border-radius: 15px;
+    font-size: smaller;
+    font-weight: 500;
+
+    &:first-of-type {
```
hunyoboy, line 9: @rsumilang This is neat.

src/plugins/ncr/components/labels-picker/index.jsx
```diff
@@ -0,0 +1,225 @@
...
+    });
+  }
+
+  handleApplyLabels() {
+    const { delegate } = this.props;
+    const { selected } = this.state;
+    delegate.handleApplyLabels(selected);
+  }
+
+  handleChange(selected) {
```
hunyoboy, line 69: nit: is it the same if you just map here? then you don't need the extra array.

` this.setState({
      selected: (selected || []).map(val => val.value),
      touched: true,
    });`

rsumilang:
@hunyoboy thanks. pr updated

src/plugins/ncr/components/labels-picker/index.jsx
```diff
@@ -0,0 +1,219 @@
...
+      selected: [],
+    };
+
+    this.isNodeDiff = this.isNodeDiff.bind(this);
+    this.handleApplyLabels = this.handleApplyLabels.bind(this);
+    this.handleCancel = this.handleCancel.bind(this);
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  UNSAFE_componentWillReceiveProps({ node }) {
```
efjacobson, line 52: @rsumilang can probably use the non-deprecated `componentDidUpdate` here

rsumilang: @efjacobson `componentDidUpdate` doesn't allow me to use setState.

gdbrown:
@rsumilang merge up master and re-test please

src/plugins/ncr/components/labels-picker/index.jsx
efjacobson: @rsumilang i think it does, you just have to be careful to not create an infinite render loop https://github.com/wb-apps/tmz-web/blob/master/client/components/gallery-widget/lightbox/index.jsx#L141

rsumilang:
@gdbrown @efjacobson pr updated and tested

src/plugins/ncr/components/labels-picker/index.jsx
```diff
@@ -49,14 +49,15 @@ class Labels extends React.PureComponent {
...
     if (!node) {
       return;
     }
     const { touched } = this.state;
     if (touched) {
       return;
     }
+    // eslint-disable-next-line react/no-did-update-set-state
     this.setState({
       selected: node.get('labels', []),
```
efjacobson, line 62: @rsumilang do you want to check if the labels have changed before re-setting them into state?

efjacobson:
@rsumilang this should go into release-next eh?

rsumilang:
@efjacobson you shouldn't see it on video nodes. this is article labels only ticket.

efjacobson:
> @efjacobson you shouldn't see it on video nodes. this is article labels only ticket.

@rsumilang galleries have it because they are labelable. by that logic, i would expect videos to have it too

rsumilang:
@efjacobson oh i see, ovp overwrites the renderSidebar thats why. Fixed.

src/plugins/ncr/components/labels-picker/delegate.js
```diff
@@ -0,0 +1,25 @@
...
+    const command = schemas.updateNodeLabels.createMessage()
+      .set('node_ref', NodeRef.fromNode(node))
+      .addToSet('add_labels', addLabels)
+      .addToSet('remove_labels', removeLabels);
+
+    dispatch(updateNodeLabels(
+      command,
+      {
+        expectedEvent: `${schemas.nodeLabelsUpdated.getCurie()}`,
+        schemas,
```
gdbrown, line 21: @rsumilang schemas config not really needed because the command/event is concrete now, meaning you configuration is needed, you know exactly what event to expect.

src/plugins/ncr/components/labels-picker/index.jsx
```diff
@@ -0,0 +1,202 @@
+import React from 'react';
+import chroma from 'chroma-js';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import isEqual from 'lodash/isEqual';
```
gdbrown, line 5: @rsumilang this should probably be called labels-form since it's nto really a picker, it just happens to have one

src/plugins/ncr/components/labels/index.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
```
gdbrown, line 1: @rsumilang i'd probably call this node-labels to it's clear it's not something like form labels or generic labels.

src/plugins/ncr/components/labels/index.jsx
```diff
@@ -0,0 +1,26 @@
...
+import PropTypes from 'prop-types';
+import chroma from 'chroma-js';
+import Message from '@gdbots/pbj/Message';
+import labelColors from 'config/labels'; // eslint-disable-line import/no-unresolved
```
gdbrown, line 5: @rsumilang call it `config/nodeLabels` for max clarity

src/plugins/ncr/components/labels/styles.scss
```diff
@@ -0,0 +1,12 @@
+span.label {
```
gdbrown, line 1: @rsumilang run the styling by matt

src/plugins/news/screens/article/schemas.js
```diff
@@ -2,6 +2,9 @@ import ArticleV1Mixin from '@triniti/schemas/triniti/news/mixin/article/ArticleV
 import SearchArticlesRequestV1Mixin from '@triniti/schemas/triniti/news/mixin/search-articles-request/SearchArticlesRequestV1Mixin';
 import resolveSchema from '@triniti/cms/utils/resolveSchema';
 
+import NodeLabelsUpdatedV1 from '@gdbots/schemas/gdbots/ncr/event/NodeLabelsUpdatedV1';
```
gdbrown, line 5: @rsumilang because these are concrete the schemas ref and there to solve the "lookup" problem we don't need it.

src/plugins/ovp/screens/video/index.jsx
```diff
@@ -29,6 +29,7 @@ class VideoScreen extends AbstractNodeScreen {
       this.renderPreviewButtons(),
       this.renderNodeStatus(),
       this.renderPublishForm(),
+      this.renderLabels(),
```
gdbrown, line 32: @rsumilang this should be LabelsForm or (since it's both displaying and allowing the user to change them)

src/plugins/ncr/components/labels-picker/index.jsx
```diff
@@ -0,0 +1,202 @@
...
+  Col,
+  Icon,
+  Row,
+} from '@triniti/admin-ui-plugin/components';
+import PicklistPicker from '@triniti/cms/plugins/sys/components/picklist-picker';
+import delegateFactory from './delegate';
+
+const getColor = (label) => labelColors[label] || '#c0c0c0';
+
+class Labels extends React.PureComponent {
```
gdbrown, line 22: @rsumilang call it LabelsForm

src/plugins/ncr/sagas/updateNodeLabelsFlow.js
```diff
@@ -0,0 +1,39 @@
...
+    getNodeRequestSchema: config.schemas.getNodeRequest,
+    pbj,
+    successMessage: `Success! The ${startCase(pbj.get('node_ref').getLabel())} was updated.`,
+    verify: (response) => {
+      if (!response.pbj.has('node')) {
+        return false;
+      }
+      const responseLabels = response.pbj.get('node').get('labels', []).sort();
+
+      // check if all add_labels was added
```
gdbrown, line 20: @rsumilang might be easier to use isInSet on the node... iterate through add/remove is do !isInSet or isInSet, no extraction/sorting/etc. needed

src/plugins/ncr/screens/node/index.jsx
```diff
@@ -320,6 +321,25 @@ export default class AbstractNodeScreen extends React.Component {
     return <Chat key="chat" nodeRef={nodeRef} />;
   }
 
+  renderLabels() {
+    const { delegate, match, isSaveDisabled } = this.props;
+    const nodeSchema = delegate.config.schemas.node || delegate.config.schemas.nodes
+      .find((schema) => schema.getCurie().getMessage().includes(match.params.type));
```
gdbrown, line 327: @rsumilang easier to just ask the node direct... 
```
const { delegate } = this.props;
    const node = delegate.getNode();
    if (!node || !node.schema.hasMixin('...')) {
       return null;
    }
    return <LabelsForm key="labels" node={node} ...blah />;
```

src/plugins/ncr/sagas/updateNodeLabelsFlow.js
rsumilang: @gdbrown not sure how thats simpler. it seems i have to loop through the results either way and check each item?

rsumilang:
@mgollnick can you please review the styles with this ticket?

rsumilang:
@gdbrown the new update has stricter line break rules:

`  28:1  error  More than 1 blank line not allowed  no-multiple-empty-lines`

I can't push without fixing them. Maybe we keep them or change the eslint rule on our end?

gdbrown:
@rsumilang disable that eslint rule, i don't agree with it.

rsumilang:
@gdbrown lint rule is fixed. i reverted most line breaks that were removed but left some because it looked more consistent with the rest of the file.

src/plugins/ncr/components/labels-form/delegate.js
```diff
@@ -0,0 +1,27 @@
...
+
+export default (dispatch, ownProps) => ({
+  handleApplyLabels: (selected) => {
+    const { schemas, node } = ownProps;
+    const origLabels = node.get('labels', []);
+    const addLabels = selected.filter((label) => !origLabels.includes(label));
+    const removeLabels = origLabels.filter((label) => !selected.includes(label));
+
+    const command = UpdateNodeLabelsV1.fromObject({
+      node_ref: NodeRef.fromNode(node),
```
gdbrown, line 14: @rsumilang minor but when doing a fromObject it expects the entire object to be plain old js object.  NodeRef.fromNode(...) returns a NodeRef instance, not a string which is typically what fromObject would get.  either use `UpdateNodeLabelsV1.create().set('node_ref', NodeRef.fromNode(node)).addToSet('labels', addLabels)...etc.` or use a string input to node_ref

src/plugins/ncr/components/labels-form/delegate.js
```diff
@@ -0,0 +1,27 @@
...
+      add_labels: addLabels,
+      remove_labels: removeLabels,
+    });
+
+    dispatch(updateNodeLabels(
+      command,
+      {
+        expectedEvent: `${schemas.nodeLabelsUpdated.getCurie()}`,
```
gdbrown, line 22: @rsumilang don't need to pass down schemas from parent(s) since it's concrete... just hard code the string here `gdbots:ncr:event:node-labels-updated`

src/plugins/ncr/screens/node/index.jsx
```diff
@@ -320,6 +321,24 @@ export default class AbstractNodeScreen extends React.Component {
...
+    if (!node || !node.schema().hasMixin('gdbots:common:mixin:labelable')) {
+      return null;
+    }
+
+    return (
+      <LabelsForm
+        disabled={!isSaveDisabled}
+        key="labels"
+        node={delegate.getNode()}
+        schemas={delegate.getSchemas()}
```
gdbrown, line 337: @rsumilang don't need to pass schemas

src/plugins/news/screens/article/schemas.js
```diff
@@ -2,6 +2,8 @@ import ArticleV1Mixin from '@triniti/schemas/triniti/news/mixin/article/ArticleV
...
```
gdbrown, line 5: @rsumilang not needed, use the static ref in the labels-form itself.  since it's concrete no mixin resolution is needed.

src/plugins/news/screens/article/schemas.js
```diff
@@ -13,6 +15,7 @@ export default {
   lockNode: resolveSchema(ArticleV1Mixin, 'command', 'lock-article'),
   markNodeAsDraft: resolveSchema(ArticleV1Mixin, 'command', 'mark-article-as-draft'),
   markNodeAsPending: resolveSchema(ArticleV1Mixin, 'command', 'mark-article-as-pending'),
+  nodeLabelsUpdated: NodeLabelsUpdatedV1.schema(),
```
gdbrown, line 18: @rsumilang not needed

src/plugins/ncr/screens/node/index.jsx
rsumilang: @gdbrown I think I need to keep the schemas in ultimately for this line: https://github.com/triniti/cms-js/blob/article-labels/src/plugins/ncr/sagas/updateNodeLabelsFlow.js#L11

gdbrown: @rsumilang that's fine, however, the additional node labels updated is not needed

rsumilang:
@gdbrown back to you

gdbrown:
@rsumilang we also need to ignore this event so user's don't get a stale data warning https://github.com/triniti/cms-js/blob/master/src/plugins/raven/services/NodeChangeWatcher.js

gdbrown:
@rsumilang please also test that labels form works on gallery, video, etc.  where ever labelable is used.

rsumilang:
@gdbrown labels on gallery and video work. back to you.

PR #245, opened 2020-08-25
add way to get from image picker to image asset page

efjacobson:


rsumilang:
@efjacobson conflict

efjacobson:
@rsumilang resolved

src/plugins/dam/components/image-asset-picker-field/index.jsx
```diff
@@ -78,12 +83,41 @@ class ImageAssetPickerField extends React.Component {
       label,
       meta: { error },
       node,
+      getNode,
     } = this.props;
 
+    const selected = getNode(value);
```
rsumilang, line 89: nit: what do you think about naming the variable to something a little more specific like `image` (or `imageNode` or something) instead of `selected`.

efjacobson:
> nit: what do you think about naming the variable to something a little more specific like image (or imageNode or something) instead of selected.

@rsumilang good call, done

PR #247, opened 2020-09-02
Force Save

hunyoboy:
What is a `force save` feature?

This is a permission-based feature where a form will force to save its current data regardless of their validity.  This is done by ignoring invalid data and persisting the valid ones. At the end of a successful submission, user will be prompted to take an action if they want to resolve the invalid data or just dismiss them.


hunyoboy:
@efjacobson please review at your convenience. thanks much.

src/plugins/ncr/sagas/forceSaveAfterSuccessFlow.js
```diff
@@ -0,0 +1,57 @@
+import { put, select } from 'redux-saga/effects';
+import { blur, getFormValues } from 'redux-form';
+import isEqual from 'lodash/isEqual';
+import swal from 'sweetalert2';
+import formData from '@triniti/cms/plugins/news/utils/formData';
+
+export default function* (config) {
```
efjacobson, line 7: @hunyoboy i  would probably name this file `restoreFormFlow`

src/plugins/ncr/sagas/forceSaveAfterSuccessFlow.js
```diff
@@ -0,0 +1,57 @@
+import { put, select } from 'redux-saga/effects';
+import { blur, getFormValues } from 'redux-form';
+import isEqual from 'lodash/isEqual';
+import swal from 'sweetalert2';
+import formData from '@triniti/cms/plugins/news/utils/formData';
```
efjacobson, line 5: @hunyoboy wouldnt want to have an ncr saga import a news util. i would probably make it a standard util in `src/utils/` and then pass it in through config as `getFormData: (key) => formData.get(key)`

src/plugins/ncr/screens/node/selector.js
```diff
@@ -51,6 +53,7 @@ export default (state, ownProps, { schemas, formName, ...rest }) => {
     || updateNodeState.status === STATUS_PENDING;
 
   const isDeleteDisabled = !isDeleteGranted || isPending || nodeStatus.toString() === 'deleted';
+  const isForceSaveDisabled = isPending;
```
efjacobson, line 56: @hunyoboy would want to account for isForceSaveGranted here, right?

src/plugins/news/services/ArticleSubscriber.js
```diff
@@ -13,6 +13,8 @@ import get from 'lodash/get';
 import isString from 'lodash/isString';
 import isUndefined from 'lodash/isUndefined';
 import { arrayPush, arrayRemoveAll, change, getFormMeta, getFormValues } from 'redux-form';
+
+import formData from '../utils/formData';
```
efjacobson, line 17: @hunyoboy would want to do an absolute import here

src/plugins/news/services/ArticleSubscriber.js
```diff
@@ -21,11 +23,18 @@ export default class ArticleSubscriber extends EventSubscriber {
...
     this.onArticleUpdated = this.onArticleUpdated.bind(this);
+    this.onClearSubmitErrors = this.onClearSubmitErrors.bind(this);
     this.onInitForm = this.onInitForm.bind(this);
     this.onSubmitForm = this.onSubmitForm.bind(this);
     this.onValidateForm = this.onValidateForm.bind(this);
   }
 
+  onClearSubmitErrors(event) {
+    const store = event.getRedux();
+    const { form } = store.getState();
```
efjacobson, line 34: @hunyoboy would want to use a selector here

efjacobson (changes requested):

src/plugins/ncr/screens/node/selector.js
hunyoboy: @efjacobson i don't think it's needed since it won't even get rendered if not granted to force save.

src/utils/formData.js
```diff
@@ -0,0 +1,20 @@
+const defaultKey = 'cms.form';
```
efjacobson, line 1: @hunyoboy not sure a default makes sense here. not much to lose by passing in `news.article`

efjacobson (changes requested):
@hunyoboy

- i noticed that the error messages dont stay after you click "resolve". can we have those? would be easier for people to see visually what is wrong

hunyoboy:
@rsumilang can you please review? thank you.

rsumilang:
@hunyoboy i lost my changes in the test environment on this article: https://localhost:3000/news/articles/a97605a8-661e-42b0-b720-9854d7ec8856/history/edit

Notice how I added the tags on "Sep 09, 2020 09:49:41 AM". The next change edited the "bar" to an invalid string and force saved it. My title change saved, my tags were cleared in the db, but I still seen the invalid title in the form.

Unfortunately it's not as easy as doing a form change on some of the fields. I had this issue with Revering History too and this is the solution we ended up with: https://github.com/triniti/cms-js/pull/228/files#diff-201f1642c4d830d96c8b064d1c18c78cR63

src/plugins/ncr/sagas/restoreFormFlow.js
```diff
@@ -0,0 +1,102 @@
...
+  }
+}
+
+function* showAlert(invalidFields) {
+  return yield swal.fire({
+    title: 'All updates have been saved, except:',
+    html: `<strong>${invalidFields.sort().join(', ')}</strong>`,
+    showCancelButton: true,
+    confirmButtonText: 'Resolve',
+    cancelButtonText: 'Dismiss',
```
rsumilang, line 48: Do we need both resolve and dismiss? Seems like they are doing the same thing.

src/plugins/ncr/utils/hasFieldUpdate.js
```diff
@@ -0,0 +1,45 @@
...
+      }
+    }
+  }
+  return null;
+};
+
+export default (formProps, invalidFields = []) => {
+  const initialValues = formProps.initialValues;
+  const values = formProps.values;
+  const registeredFields = formProps.registeredFields || {};
```
rsumilang, line 34: @hunyoboy this can be:

`const { initialValues, registeredFields = {}, values } = formProps;`

hunyoboy: @rsumilang definitely thanks. updated.

rsumilang:
@hunyoboy looks good

rsumilang (approved):

PR #249, opened 2020-09-11
v0.10.0

bhsiao55:
* Create permission based force-save button for articles.
* Adding support for `recommendations_enabled` field.
* Use `@triniti/schemas` v1.1.12


bhsiao55:
Create permission based force-save button for articles has been rejected

PR #251, opened 2020-09-21
Blocksmith Invalid State Prevention

hunyoboy:


hunyoboy:
@efjacobson please review when you have a chance. thank you.

efjacobson:
@hunyoboy i played around with this a bit and found some other edge cases, let me know what you think of this: https://github.com/triniti/cms-js/pull/253

hunyoboy:
> @hunyoboy i played around with this a bit and found some other edge cases, let me know what you think of this: #253
> 
> btw
> 
> ```
> `<p><figure>${text}</figure></p>`
> ```
> 
> is not valid for a text block, there should not be any figure tag, should be
> 
> ```
> `<p>${text}</p>`
> ```

thank you for pointing out and i appreciate. with this, i found an opportunity that this tag can possibly be pasted by a user. this could potentially be the root cause. let me know.

efjacobson (commented):
@hunyoboy we lost the error alert?

hunyoboy:
> @hunyoboy we lost the error alert? also testing this im not sure what value we get out of the safe saga - what do you think it brings to the table?

@efjacobson the main benefit is if a new and similar issue happens again, we can confidently say that the cause is NOT an uncatch error on the `updateNodeFlow` saga. By this, it can considerably lessen our surface area to debug at the same time, increase our efficiency in finding the issue. however if we’re sure that `blocksmith invalid state` is the sole culprit of the HL issue and we’re just aiming to fix it, i would agree to remove it. also, we can even just leave those error alerts as ease ( like prod ) but of course fix the small bugs we noticed ( on news subscriber ) leaving the PR much much slimmer. let me know.

efjacobson:
@hunyoboy yeah, i agree, lets just fix the copy-paste but and article subscriber and call it a day

efjacobson (approved):
nice work @hunyoboy

PR #264, opened 2020-11-24
Blocksmith error reporting & recovery

efjacobson:
- removed a bunch of unwise exceptions thrown for Blocksmith jazz. they should never have been exceptions and would pollute the error reporting that is coming soon
- setState functions for all the blocksmith things
- editorState argument preference when applicable (https://draftjs.org/docs/advanced-topics-editorstate-race-conditions/)
- avoid some use of `selectBlock` for atomic blocks - this is a vestigial holdover from when you could select atomic blocks (before the focus plugin was removed)
- whenever EditorState.push would have been used _to get a new EditorState that would immediately be passed into the Editor component_, now use the `pushEditorState` util, which has validation baked in
- if the blocks would be invalid on saving, the article screen's delegate now checks for that and will display an alert. the options are "dont save" or "save only valid blocks" (this could be added to page/person which also uses the editor but it seems unnecessary)
- 3 new error boundaries: one for the Editor, one around the block placeholders, and one around the Modals
- if something causes the Blocksmith error boundary to throw and the blocks are actually valid, attempt to restore. if at least one of the blocks is not valid, show a simple read-only version of the contents and buttons for restoring all or only invalid blocks
- if the error boundary throws enough times (`MAX_ERROR_COUNT`), even if all the blocks are valid, show the simple read-only version of the contents
- you can simulate an ErrorBoundary condition by just throwing in the render of one of the ErrorBoundary components or one of its children.
- you can simulate an invalid block by just throwing in `validateBlocks` (during testing i just threw for any video-block)
- you can simulate a modal ErrorBoundary in a non-hacky way by pasting a bunch of text (longer than the schema field allows) into a block modal. an article block's `link_text`, for example, will throw if you paste in more than 255 characters
- you can simulate an invalid "text" block by undoing Joel's fix for that https://github.com/triniti/cms-js/commit/4daa8bb422a3a041fbd926c0f14de4e04cb1010a and pasting copy/pasting an image in addition to text


efjacobson:
@arunkarnati lmk if you want to go over some of this together, i can give a little demo

src/plugins/blocksmith/components/modal-error-boundary/index.jsx
```diff
@@ -0,0 +1,47 @@
...
+    };
+  }
+
+  static getDerivedStateFromError() {
+    return { hasError: true };
+  }
+
+  componentDidCatch(error, errorInfo) {
+    const { onCloseModal } = this.props;
+    console.error(error, errorInfo);
```
arunkarnati, line 24: @efjacobson - is this intentional?

src/plugins/blocksmith/resolver.js
```diff
@@ -287,13 +287,13 @@ export const getPlaceholder = (type, props = {}) => {
       component = createLazyComponent(import('@triniti/cms/plugins/blocksmith/components/youtube-video-block-placeholder'));
       break;
     default:
-      return null;
+      component = createLazyComponent(import('@triniti/cms/plugins/blocksmith/components/default-block-placeholder'));
```
arunkarnati, line 290: @efjacobson - when does this default-block gets returned?

src/plugins/news/screens/article/delegate.js
```diff
@@ -9,6 +13,31 @@ class Delegate extends AbstractDelegate {
       formName: formNames.ARTICLE,
     }, dependencies);
   }
+
+  handleSubmit(data, formDispatch, formProps) { // eslint-disable-line consistent-return
+    const { blocksmithState, dispatch } = this.component.props;
+    const { isValid, validEditorState } = validateBlocks(blocksmithState.editorState);
+
+    if (isValid) {
+      return AbstractDelegate.prototype.handleSubmit.call(this, data, formDispatch, formProps);
```
arunkarnati, line 22: How about calling the parent method like this?

```suggestion
      return super.handleSubmit(data, formDispatch, formProps);
```

src/plugins/news/screens/article/delegate.js
```diff
@@ -9,6 +13,31 @@ class Delegate extends AbstractDelegate {
...
+    swal.fire({
+      title: 'Error saving article - one or more blocks is invalid.',
+      text: 'The invalid blocks cannot be saved. Would you like to save only the valid blocks?',
+      type: 'warning',
+      showCancelButton: true,
+      confirmButtonText: 'Yes',
+      confirmButtonClass: 'btn btn-danger',
+      cancelButtonClass: 'btn btn-secondary',
+    }).then((result) => {
+      if (!result.value) {
```
arunkarnati, line 34: @efjacobson - do we even need this check here? I mean, we can handle the `if (result.value)` case.

```
if (result.value) {
  dispatch(storeEditor(formProps.form, validEditorState));
  dispatch(submit(formProps.form));
}
```

arunkarnati (changes requested):
@efjacobson - looks good overall. I left a few comments/suggestions.

Also, I suggest adding details about error boundaries to the blocksmith README file under the FYI section. https://github.com/triniti/cms-js/blob/5bcd11804a7f9561b5a05946505beae199dfc7b2/src/plugins/blocksmith/components/blocksmith/README.md. You can copy and paste a few bullet points straight from the PR description.

src/plugins/blocksmith/components/blocksmith/index.jsx
```diff
@@ -153,32 +159,36 @@ class Blocksmith extends React.Component {
...
       blocksmithState,
       delegate,
-      formName,
       isEditMode,
       node,
     } = props;
 
     const decorator = new MultiDecorator([new CompositeDecorator(decorators)]);
     let editorState = EditorState.createEmpty(decorator);
+    let errors = {};
```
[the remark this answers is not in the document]
efjacobson: @arunkarnati its a lint thing

Screen Shot 2020-12-01 at 1 58 05 PM

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # CHANGELOG
 
 
+## vN.N.N
+* Error reporting & recovery
```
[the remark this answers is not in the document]
efjacobson: done

src/plugins/news/screens/article/delegate.js
efjacobson: @arunkarnati works for me, updated

src/plugins/news/screens/article/delegate.js
efjacobson: @arunkarnati good call, updated

src/plugins/blocksmith/resolver.js
efjacobson: @arunkarnati if something like what Joel fixed happens again (the copy paste with image). where a block is considered to be atomic by draft, but `message` is undefined here https://github.com/triniti/cms-js/blob/error-reporting-recovery/src/plugins/blocksmith/components/blocksmith/index.jsx#L608

src/plugins/blocksmith/components/modal-error-boundary/index.jsx
efjacobson: @arunkarnati i suppose its not necessary, it will get tied into the api error reporting eventually. removed for now

src/plugins/blocksmith/components/blocksmith/index.jsx
```diff
@@ -291,16 +307,36 @@ class Blocksmith extends React.Component {
...
-    const { isEditMode } = this.props;
+  componentDidUpdate({ editorState: prevPropsEditorState, isEditMode: prevIsEditMode }) {
+    const { editorState: currentPropsEditorState, isEditMode } = this.props;
+    const { editorState } = this.state;
     if (prevIsEditMode !== isEditMode) {
       // eslint-disable-next-line react/no-did-update-set-state
       this.setState(() => ({ readOnly: !isEditMode }));
+      return;
+    }
+    if (
```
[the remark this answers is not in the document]
efjacobson: @arunkarnati currentPropsEditorState is used right after this condition. if you do it your way, try saving and you will see that it crashes after

src/plugins/blocksmith/components/blocksmith/error-boundary/index.jsx
```diff
@@ -0,0 +1,169 @@
...
+            }
+            return (
+              <div className="preview-component preview-component__error" key={block.getKey()}>
+                <p><strong>{`invalid ${message}`}</strong></p>
+                <p>{block.toString()}</p>
+              </div>
+            );
+          }
+          const message = block.schema().getId().getCurie().getMessage();
+          let Component = null;
```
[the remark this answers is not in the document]
efjacobson: @arunkarnati another lint thing

Screen Shot 2020-12-01 at 2 05 36 PM

src/plugins/blocksmith/components/blocksmith/error-boundary/index.jsx
```diff
@@ -0,0 +1,169 @@
...
+        {!isValid && (
+          <p>The invalid blocks are styled in <span className="invalid-block-indicator"><strong>red</strong></span> below.</p>
+        )}
+        <hr />
+        {blocks.map(({ block }) => {
+          if (block instanceof ContentBlock) {
+            let message = 'block';
+            const blockData = block.getData();
+            if (blockData && blockData.has('canvasBlock')) {
+              message = blockData.get('canvasBlock').schema().getId().getCurie().getMessage(); // eslint-disable-line
```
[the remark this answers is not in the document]
efjacobson: @arunkarnati not if something like what Joel fixed comes back - you can test it yourself by doing this thing you recommended, reverting Joel's fix, and then pasting in that mutant payload from the nick cannon article. you will see that it crashes the editor. this gets caught by the ErrorBoundary but imo it is better to display the invalid block placeholder

src/plugins/blocksmith/components/blocksmith/index.jsx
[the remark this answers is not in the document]
arunkarnati: @efjacobson - I meant to say to leave the declaration but remove initialization. `let errors;`

src/plugins/blocksmith/components/blocksmith/error-boundary/index.jsx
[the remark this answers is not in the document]
arunkarnati: @efjacobson - I meant to say to leave the declaration but remove the initialization. `let Component;`

src/plugins/blocksmith/resolver.js
arunkarnati: @efjacobson - IC, in the edge case you mentioned above, what will be the value of `blockData.get('canvasBlock').schema()`? `undefined`?

efjacobson: @arunkarnati in this edge case there will be an `atomic` block that does not have a `canvasBlock` in its data. so yes, the `type` parameter for this function will be `undefined`.

efjacobson: @arunkarnati in this edge case `blockData.get('canvasBlock').schema()` will throw

src/plugins/blocksmith/components/blocksmith/error-boundary/index.jsx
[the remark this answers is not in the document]
efjacobson: @arunkarnati gotcha - updated

src/plugins/blocksmith/components/blocksmith/index.jsx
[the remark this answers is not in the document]
efjacobson: @arunkarnati gotcha - updated

src/plugins/blocksmith/components/blocksmith/error-boundary/index.jsx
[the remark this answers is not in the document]
efjacobson: @arunkarnati oic what you mean, fix incoming

src/plugins/blocksmith/resolver.js
arunkarnati: @efjacobson - got it, thanks!

src/plugins/blocksmith/components/blocksmith/error-boundary/index.jsx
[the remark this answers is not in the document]
efjacobson: @arunkarnati updated here and a few other places

arunkarnati (approved):

arunkarnati:
@efjacobson - one last thing, are you planning to merge this into the master branch?

efjacobson:
@gdbrown over to you

gdbrown (approved):

PR #268, opened 2020-12-04
fix for: Raven can get caught in an alert loop

efjacobson:


efjacobson:
@gdbrown lmk what you think now

src/plugins/raven/sagas/connectionFlow.js
```diff
@@ -88,6 +77,16 @@ export default function* (raven) {
       yield take(connectsChannel);
       attempts = 0;
     } catch (e) {
+      yield swal.fire({
```
gdbrown, line 80: @efjacobson i would not alert until after 10 attempts

gdbrown (changes requested):

src/plugins/raven/sagas/connectionFlow.js
efjacobson: done

efjacobson:
@gdbrown seems to me that the way it is now, after the 10th failed attempt, the swal will fire and after that nothing will happen as it will just continue forevermore

gdbrown (approved):

PR #271, opened 2020-12-16
Blocksmith crash when text block contains only carriage return

efjacobson:


src/plugins/blocksmith/utils/validateBlocks.js
```diff
@@ -19,20 +19,32 @@ export default (editorState) => {
...
+          error: 'Error: unable to convert to canvas block.',
+        };
+      }
     } catch (e) {
       isValid = false;
       console.error(`[ERROR:Blocksmith:util:validateBlocks] - Block: ${block.toString()} - ${e}`);
       blocks.push({
         block,
-        error: e.stack,
+        error: e.stack || e.toString(),
```
gdbrown, line 43: @efjacobson i'd probably template string that instead `${e}`

src/plugins/blocksmith/utils/convertToCanvasBlocks.js
```diff
@@ -37,7 +37,10 @@ export default (editorState, allowEmptyTextBlocks = false) => {
     blockRenderers: {
       [blockTypes.ATOMIC]: (block) => `${CANVAS_BLOCK_TOKEN}${JSON.stringify(block.getData().get('canvasBlock'))}`,
       [blockTypes.UNSTYLED]: (block) => {
-        if (allowEmptyTextBlocks && block.getText() === '') {
+        if (
+          allowEmptyTextBlocks
```
gdbrown, line 41: @efjacobson maybe trim the block.getText() and look for empty string intead so you don't need regex at all

gdbrown (changes requested):

gdbrown (approved):

PR #272, opened 2020-12-16
Fix Login Flow Issue

hunyoboy:


hunyoboy:
@efjacobson please review if you're free. thank you.

src/plugins/iam/sagas/index.js
```diff
@@ -10,15 +10,20 @@ import { actionTypes, serviceIds } from '../constants';
...
  */
 function* watchLogin(container) {
+  let lastTask;
   while (true) {
     const { doLogin } = yield race({
       doLogin: take(actionTypes.LOGIN_REQUESTED),
       doLogout: take(uiActionTypes.LOGOUT_REQUESTED),
     });
 
+    if (lastTask) {
```
efjacobson, line 20: @hunyoboy might call this `loginTask` and only check to see if we need to cancel it in the `if (doLogin)`block

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # CHANGELOG
 
 
+## vN.N.N
+* issue #1238: cms will crash after session/raven disconnect and users accept new disconnect Alert
```
efjacobson, line 5: @hunyoboy remember we dont link to tickets that are not part of this repo

efjacobson (changes requested):

CHANGELOG.md
hunyoboy: @efjacobson removed the link and updated changed log.

src/plugins/iam/sagas/index.js
hunyoboy: @efjacobson thank you i got you but i think we also need to cancel if for some reason a logout request will be issued. if for some reason the login flow is still running ( auth server still pending, etc. ) and suddenly we want to logout, i think we really need to terminate the login flow to keep the state in sync.

efjacobson:
this looks good to me, what do you think @gdbrown ?

efjacobson (approved):

gdbrown (changes requested):
@hunyoboy can we also put a try catch around the raven connect and simply return if it fails.

https://github.com/triniti/cms-js/blob/1238-login-flow-issue/src/plugins/raven/services/Raven.js#L79

i believe that was also causing the white screen.

gdbrown (approved):

PR #274, opened 2020-12-17
Set user node title using first and last name fields

arunkarnati:


src/plugins/iam/services/UserSubscriber.js
```diff
@@ -89,6 +89,7 @@ export default class UserSubscriber extends EventSubscriber {
     node.set('email', data.email)
       .set('first_name', data.firstName)
       .set('last_name', data.lastName)
+      .set('title', `${data.firstName} ${data.lastName}`)
```
efjacobson, line 92: @arunkarnati 

https://github.com/gdbots/iam-php/blob/36113916a034f8b205a7055c07fde0703b9b3a2b/src/CreateUserHandler.php#L47
and in the next version
https://github.com/gdbots/iam-php/blob/master/src/UserAggregate.php#L146

arunkarnati: @efjacobson - well, that is the problem here:

when updating a node, the _newNode_ will have the _oldNode's_ title unless explicitly set in the _UserSubscriber_

https://github.com/gdbots/iam-php/blob/v1.0.0/src/UpdateUserHandler.php#L54,
https://github.com/gdbots/iam-php/blob/master/src/UserAggregate.php#L139

It's easy to reproduce, update the first/last name of a user, and see if the title value changes.

efjacobson: @arunkarnati i get it, i just feel like if we wanted title to be set by the cms we would have a title field

arunkarnati: @efjacobson - Currently, we have instances where we set values to fields without having an actual form field in the UI, here's one such example: https://github.com/triniti/cms-js/blob/master/src/plugins/notify/services/NotificationEnricher.js#L24

I'm trying to understand why setting a title in the subscriber would be a problem when we know that the server-side code ignores if it is already set, and we know it fixes the inconsistency.

efjacobson: @arunkarnati well, it makes the server code useless

arunkarnati: @efjacobson - I see your point. I updated the code to set the title when `!isCreateForm`. 

That way, the `setDefaultTitle` fn on the server-side will still get used when the node is created, and the same fn call when a node is updated is useless right now anyway since it won't even get to this line: https://github.com/gdbots/iam-php/blob/master/src/UserAggregate.php#L143

efjacobson (approved):

PR #276, opened 2020-12-22
Active-Edits To Only Refresh On Click

hunyoboy:


hunyoboy:
@gdbrown please review if you have a minute. thank you.

src/plugins/raven/components/active-edits-table/index.jsx
```diff
@@ -16,84 +16,70 @@ import delegate from './delegate';
...
+  const handleRefresh = async () => {
+    setStatus(STATUS_PENDING);
+    await handleUpdateCollaborations(accessToken);
+    setStatus(STATUS_FULFILLED);
+  };
 
-    static defaultProps = {
-      title: 'Active Edits',
-    }
+  useEffect(() => {
```
gdbrown, line 32: @hunyoboy `useEffect(handleRefresh, []);` is all you need, arrow function isn't needed.

src/plugins/raven/components/active-edits-table/index.jsx
```diff
@@ -16,84 +16,70 @@ import delegate from './delegate';
...
+    handleRefresh();
+  }, []);
 
-    constructor(props) {
-      super(props);
-      this.handleRefresh = this.handleRefresh.bind(this);
-      this.state = { intervalId: null, status: null };
-    }
+  const memoizedTableRows = useMemo(() => collaborationNodes.map((node, idx) => (
```
gdbrown, line 36: @hunyoboy memoized isn't really needed here, optimized in this area is not important.

gdbrown (changes requested):

src/plugins/raven/components/active-edits-table/index.jsx
hunyoboy: @gdbrown i tried that format and it breaks on 'unmount'.

gdbrown: @hunyoboy okee dokee, ic, because you're using an async function which returns a value and react is trying to call it.  ok, that's fine

src/plugins/raven/components/active-edits-table/index.jsx
hunyoboy: @gdbrown yes the purpose of this memo is so the TableRows won't get auto-refresh when new collaborated nodes are added/removed and detected by Raven. i thought it won't but i noticed Raven will still auto-refresh the table. with this, it will keep using the memoized TableRow(s) until a change in status due to button click occurs.

gdbrown (approved):

PR #281, opened 2021-01-08
Google map center field

bhsiao55:
* Set center field on Google Map Block using Geocoding API.


src/plugins/blocksmith/components/google-map-block-modal/index.jsx
```diff
@@ -74,24 +81,39 @@ export default class GoogleMapBlockModal extends React.Component {
...
     onAddBlock(this.setBlock());
     toggle();
   }
 
-  handleEditBlock() {
+  async handleEditBlock() {
+    const { oldQ, q, center } = this.state;
     const { onEditBlock, toggle } = this.props;
+
+    if (q !== oldQ || !center) {
```
arunkarnati, line 111: @bhsiao55 - if the _oldQ_ is only used to compare with the initial value, then let's remove it and get that value directly from the block like this:

```
const { q, center } = this.state;
const { block, onEditBlock, toggle } = this.props;

if (q !== block.get('q') || !center) {
```

src/plugins/blocksmith/components/google-map-block-modal/getGoogleMapCoordinates.js
```diff
@@ -0,0 +1,26 @@
...
+  return new Promise((resolve) => {
+    const status = data.status;
+    if (status !== 'OK' || data.results.length === 0) {
+      console.error('Geocoding failed:', status);
+      resolve(null);
+      return;
+    }
+
+    const results = data.results[0];
+    resolve(`${results.geometry.location.lat},${results.geometry.location.lng}`);
```
arunkarnati, line 22: @bhsiao55 - move this line to _try_ block, so the exception is caught if the Google Maps API response doesn't have any results.

src/plugins/blocksmith/components/google-map-block-modal/getGoogleMapCoordinates.js
```diff
@@ -0,0 +1,26 @@
+/* globals GOOGLE_MAPS_API_KEY */
+const getGoogleMapCoordinates = async (str) => {
```
arunkarnati, line 2: @bhsiao55 - how does `getCoordinatesFromAddress` sound for the file name? We don't have to know what service we are using by looking at the file name.

arunkarnati (changes requested):

src/plugins/blocksmith/components/google-map-block-preview/index.jsx
```diff
@@ -11,7 +11,7 @@ const GoogleMapBlockPreview = ({ block, width }) => {
       title="google-map-block-preview"
       width={width}
       height={width / (16 / 9)}
-      src={`https://www.google.com/maps/embed/v1/place?key=${GOOGLE_MAPS_API_KEY}${zoomParam}&maptype=${block.get('maptype') || 'roadmap'}&q=${block.get('q') || ''}${block.has('center') ? `&center=${block.get('center').getLatitude()},${block.get('center').getLongitude()}` : ''}`}
+      src={`https://www.google.com/maps/embed/v1/place?key=${GOOGLE_MAPS_API_KEY}${zoomParam}&maptype=${block.get('maptype') || 'roadmap'}&q=${block.get('q') || ''}`}
```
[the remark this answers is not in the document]
bhsiao55: if both is there center takes precedence.  I think it should be removed cause on web we don't even use it

src/plugins/blocksmith/components/google-map-block-modal/getGoogleMapCoordinates.js
```diff
@@ -0,0 +1,26 @@
...
+  let data = {};
+
+  try {
+    const response = await fetch(url);
+    data = await response.json();
+  } catch (error) {
+    data.status = error;
+  }
+
+  return new Promise((resolve) => {
```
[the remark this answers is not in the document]
bhsiao55: removed explicit promise

src/plugins/blocksmith/components/google-map-block-preview/index.jsx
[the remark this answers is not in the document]
arunkarnati: @bhsiao55 - if the map is going to be more precise when the *center* field is part of the URL, then let's leave it in there. Otherwise, please remove it.

src/plugins/blocksmith/components/google-map-block-modal/getGoogleMapCoordinates.js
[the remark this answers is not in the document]
bhsiao55: added

src/plugins/blocksmith/components/google-map-block-modal/getGoogleMapCoordinates.js
bhsiao55: yup makes sense

src/plugins/blocksmith/components/google-map-block-modal/getGoogleMapCoordinates.js
bhsiao55: file name changed

src/plugins/blocksmith/components/google-map-block-modal/index.jsx
bhsiao55: yes that works out better

arunkarnati (approved):
@bhsiao55 - I made some very minor updates to the util. Please test it before merging. Otherwise, looks good to me.

arunkarnati (approved):

PR #283, opened 2021-01-13
v0.12.0

gdbrown:
* Reconnect raven on every `@triniti/raven/PUBLISH_MESSAGE_REQUESTED`.
* Update raven mqtt to disable reconnect until transformWsUrl supports promises.
* Use toast alerts for stale data when collaborating


src/plugins/ncr/screens/node/AbstractDelegate.js
```diff
@@ -546,7 +547,9 @@ export default class AbstractDelegate {
       if (!document.hidden) {
         const node = this.getNode();
         this.dispatch(sendHeartbeat(`${nodeRef}`, node ? node.get('etag') : null));
-        this.startCollaborationMonitor();
+        if (isEditMode) {
```
efjacobson, line 550: @gdbrown i think i should have removed this when i reverted the position, eh? since we know isEditMode is true because of `if (!isEditMode || !canCollaborate || !isCollaborating) {`

gdbrown: @efjacobson yes, revert all AbstractDelegate changes

efjacobson (approved):

PR #284, opened 2021-01-15
select poster image as title instead of select primary image

bhsiao55:
* Add `modalTitle` prop to image asset picker modal and image asset picker field.

### DIFF
--- a/src/plugins/dam/components/image-asset-picker-modal/index.jsx
+++ b/src/plugins/dam/components/image-asset-picker-modal/index.jsx
@@ -29,7 +29,7 @@ class ImageAssetPickerModal extends React.Component {
     galleryNode: PropTypes.instanceOf(Message),
     isOpen: PropTypes.bool,
     label: PropTypes.string,
-    modalTitle: PropTypes.string,
+    title: PropTypes.string,
     multiAssetErrorMessage: PropTypes.string,
     node: PropTypes.instanceOf(Message),
     onCloseUploader: PropTypes.func,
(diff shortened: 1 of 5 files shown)


src/plugins/dam/components/image-asset-picker-modal/index.jsx
```diff
@@ -29,6 +29,7 @@ class ImageAssetPickerModal extends React.Component {
     galleryNode: PropTypes.instanceOf(Message),
     isOpen: PropTypes.bool,
     label: PropTypes.string,
+    modalTitle: PropTypes.string,
```
hunyoboy, line 32: @bhsiao55 since you're already in the modal, you can just call this title.

hunyoboy (commented):
@bhsiao55 looks good. just a comment.

src/plugins/dam/components/image-asset-picker-modal/index.jsx
```diff
@@ -29,6 +29,7 @@ class ImageAssetPickerModal extends React.Component {
     galleryNode: PropTypes.instanceOf(Message),
     isOpen: PropTypes.bool,
     label: PropTypes.string,
```
hunyoboy, line 31: @bhsiao55 i don't think you need the label props now really! you can just default your 'title' to current label's value.

src/plugins/dam/components/image-asset-picker-modal/index.jsx
bhsiao55: yup will change

src/plugins/dam/components/image-asset-picker-modal/index.jsx
bhsiao55: label still being used by other components

hunyoboy (approved):

PR #285, opened 2021-01-22
Cms errors logstream

cohencodes:


efjacobson (changes requested):
@cohencodes 

- this would be a good error to know about https://github.com/triniti/cms-js/blob/master/src/plugins/blocksmith/utils/validateBlocks.js#L38
- and all the collaboration/collection ones

i also think we should prevent this from happening locally, otherwise the dev/test streams are going to be polluted by people working on tickets. what do you think @gdbrown ?

src/plugins/raven/services/Raven.js
```diff
@@ -201,6 +208,30 @@ export default class Raven {
...
+  onError(error) {
+    const state = this.store.getState();
+    const accessToken = getAccessToken(state);
+
+    const logData = {
+      app_version: APP_VERSION,
+      error: JSON.stringify(error, Object.getOwnPropertyNames(error)).replaceAll('://', '[PROTOCOL_TOKEN]').replace(/(\/\.\.)+\/?/g, '[UP_DIRECTORY_TOKEN]'),
+    };
+
+    if (!isJwtExpired(accessToken)) {
```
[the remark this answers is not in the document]
gdbrown: @efjacobson @cohencodes i agree, logging when running locally is probably not needed.

cohencodes:
@gdbrown @efjacobson changes made

efjacobson:
@cohencodes how about this for the oddness of not knowing which argument is the actual Error object

```
onError(...args) {

...

const error = args.find(arg => arg instanceof Error) || args[0]
```

because sometimes the first argument is the error message which means the stack trace wont be in the payload

cohencodes:
@efjacobson I like that, going to test it all again locally to make sure everything's kosher

cohencodes:
@efjacobson i couldn't recreate any instance of the stack trace not being available, however i added that change and it works after testing, i don't think it hurts to add that.

the only thing i'm still iffy about is how many times we log the same error when its a react error, that part seems a bit funky to me but if you guys think its cool then this should be gtg now

efjacobson:
@cohencodes fyi an example of when the first argument will not be an `Error` is if you were to throw in `onInitForm` in a subscriber. without the new code it would just be the error message, no stack

cohencodes:
okay got it @efjacobson i've made all the updates, including only logging react errors once

efjacobson:
@cohencodes i think you can just remove the `window.onerror` in the error boundaries. its a bit murky how those bubble but that looks like it works

efjacobson:
@cohencodes this is also log worthy https://github.com/triniti/cms-js/blob/master/src/plugins/raven/sagas/connectionFlow.js#L53

src/plugins/raven/services/Raven.js
```diff
@@ -201,6 +209,39 @@ export default class Raven {
...
+   *
+   */
+  onError(...args) {
+    const state = this.store.getState();
+    const accessToken = getAccessToken(state);
+
+    if (
+      isJwtExpired(accessToken)
+      || window.location.hostname === 'localhost'
+      || this.logStreamRequestCount > 1
```
efjacobson, line 223: @cohencodes do you still need this logStreamRequestCount check?

src/plugins/raven/services/Raven.js
```diff
@@ -41,6 +43,8 @@ export default class Raven {
     this.mqtt = mqtt;
     this.store = store;
     this.apiEndpoint = apiEndpoint;
+    this.logStreamRequestCount = 0;
+    window.onerror = this.onError.bind(this);
```
efjacobson, line 47: @cohencodes nit: i would put this at the end of the constructor instead of in the middle of all the `this` assignments

efjacobson (changes requested):

src/plugins/raven/services/Raven.js
cohencodes: @efjacobson this is still needed otherwise we get multiple logs for the same error, removing from error boundary didn't resolve that

src/plugins/raven/services/Raven.js
cohencodes: fixed

src/plugins/raven/services/Raven.js
efjacobson: @cohencodes yeah but i think just not logging the first error that happens is tricky - what if the first error is not one that would be double reported?

how about keeping track of the md5 hashes of the stringified payload and, if that hash already exists, dont send again

cohencodes: okay i will try this

src/plugins/raven/services/Raven.js
```diff
@@ -201,6 +211,39 @@ export default class Raven {
...
 
+  /**
+   * @link https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror
+   *
+   */
+  onError(...args) {
+    const state = this.store.getState();
+    const accessToken = getAccessToken(state);
+    const error = args.find((arg) => arg instanceof Error) || args[0];
+    const hashName = md5(error);
```
efjacobson, line 222: @cohencodes hash the stringified error

src/plugins/raven/services/Raven.js
```diff
@@ -201,6 +211,39 @@ export default class Raven {
...
+    const logData = {
+      app_version: APP_VERSION,
+      error: JSON.stringify(error, Object.getOwnPropertyNames(error)).replaceAll('://', '[PROTOCOL_TOKEN]').replace(/(\/\.\.)+\/?/g, '[UP_DIRECTORY_TOKEN]'),
+    };
+
+    if (
+      isJwtExpired(accessToken)
+      || window.location.hostname === 'localhost'
+      || hashName === this.hashName
```
efjacobson, line 231: @cohencodes i would call these `errorHash` and `this.prevErrorHash`

efjacobson (changes requested):

efjacobson (approved):

efjacobson:
@cohencodes let Greg take a look at this before you merge

src/plugins/raven/services/Raven.js
```diff
@@ -201,6 +210,39 @@ export default class Raven {
...
+  /**
+   * @link https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror
+   *
+   */
+  onError(...args) {
+    const state = this.store.getState();
+    const accessToken = getAccessToken(state);
+    const error = args.find((arg) => arg instanceof Error) || args[0];
+    const logData = {
+      app_version: APP_VERSION,
```
gdbrown, line 222: @cohencodes also include the current location... like `request_uri: window.location`

gdbrown (changes requested):

src/plugins/raven/services/Raven.js
cohencodes: getting a cors error when trying to send this, need to mess around with it more

efjacobson: @cohencodes probably need to replace the `://` here too

cohencodes: @efjacobson oh yeah i already tried that it still doesn't like it though

gdbrown: @cohencodes if we're ignoring localhost later why do we need to do this replacement at all?

cohencodes: @gdbrown i was thinking that too, technically we don't need it, I can remove

gdbrown (approved):

PR #294, opened 2021-02-24
Non image asset upload

efjacobson:


src/plugins/dam/sagas/uploadFileFlow.js
```diff
@@ -47,32 +47,18 @@ export default function* (hashName, fileInfo, fileBuffer, preSignedUrl, variant)
...
-        } else {
-          const assetResponse = yield call(fetch, previewUrl, { method: 'get' });
-          if (assetResponse.status !== 200) {
-            throw new Error('dam upload file flow failed first try. trying again.');
-          }
-        }
-      } catch (e) {
-        // Make one more try after 2 secs because sometimes amazon
-        // lags on making the asset available... wtf amazonians
-        yield delay(2000);
```
rsumilang, line 67: @efjacobson this delay was kind've important because sometimes it took a moment for an image preview to become available. without it, we could go over the MAX_RETRIES and possibly alert a user that the image failed even though it was possibly successful.

efjacobson: @rsumilang like, if checkImage rejects? or do you mean the presigned PUT. it does still have the delay at the beginning of the loop if it is not the first retry

rsumilang: @efjacobson the `put` was usually good but checking if an image was available for previewing via `checkImage` had a high fail rate on the initial check as it took amazon a while to make asset available via url. I'd feel more comfortable with this change if we had a higher `MAX_RETRIES` for images.

efjacobson: @rsumilang ok yeah i can whip something up. think i'll only retry the PUT if it fails though, seems odd to do that again if it was just the checkImage that failed

efjacobson: @rsumilang lmk what you think now. the delay is now 2 seconds per retry, max of 5, and i wrapped the `yield put(e)` in its own try since it is `put` is only supposed to be used for plain actions and was choking on exceptions

src/plugins/dam/sagas/uploadFileFlow.js
```diff
@@ -28,51 +28,37 @@ export default function* (hashName, fileInfo, fileBuffer, preSignedUrl, variant)
...
   }
 
+  let uploadSucceeded = false;
   for (let retries = 0; retries < MAX_RETRIES; retries += 1) {
     try {
-      // Delay the process by 2 seconds if this isn't the first try
-      if (retries) {
-        yield delay(2000);
-      }
+      yield delay(retries * 2000);
```
rsumilang, line 34: why are the 2 second delays being multiplied?

efjacobson: @rsumilang simple way to have no delay at first and then an increasing delay after failures. is your concern that a total failure will happen after too long of a delay?

rsumilang: @efjacobson I think further increasing delay times is unnecessary. Increasing the MAX_RETRIES should be sufficient.

efjacobson: @rsumilang updated

rsumilang:
@efjacobson 👍

PR #302, opened 2021-04-06
Uploader Component Fixes

hunyoboy:
1. fix issue when user tries to update a file while it is being uploaded.
2. fix issue when user tries to update a file that has an upload error.
3. limit update/patch process to completely uploaded files ONLY.
4. exactly compare count of patched files to Raven's dispatched 'asset-patched' action.
5. fix warning that occurs when clicking for the first time the credit's `apply to all` button.

### DIFF
--- a/src/plugins/dam/components/uploader/delegate.js
+++ b/src/plugins/dam/components/uploader/delegate.js
@@ -173,17 +173,17 @@ class Delegate extends AbstractDelegate {
   }
 
   handleCreditApplyAll(credit, files, initialValues) {
-    const assetCount = { assetCount: files.length };
+    const config = { ...this.config, assetCount: Object.keys(files).length };
     const data = {
       fields: ['credit'],
       values: { credit },
     };
     this.dispatch(initialize(this.getFormName(), { ...initialValues, credit }, 'credit'));
-    this.dispatch(patchAssets(data, files, { ...this.config, assetCount }));
+    this.dispatch(patchAssets(data, files, config));
   }
 
   handleExpiresAtApplyAll(expiresAt, files, initialValues) {
-    const assetCount = { assetCount: files.length };
+    const config = { ...this.config, assetCount: Object.keys(files).length };
     const data = {
       fields: ['expires_at'],
       values: { expires_at: expiresAt },
@@ -192,7 +192,7 @@ class Delegate extends AbstractDelegate {
       this.getFormName(),
       { ...initialValues, expiresAt },
     ));
-    this.dispatch(patchAssets(data, files, { ...this.config, assetCount }));
+    this.dispatch(patchAssets(data, files, config));
   }
 
   getFormName() {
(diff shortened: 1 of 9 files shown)


hunyoboy:
@efjacobson please review if you have chance. thank you.

src/plugins/dam/sagas/patchAssetsFlow.js
```diff
@@ -45,13 +45,17 @@ export default function* (action) {
   const expectedEvent = config.schemas.assetPatched.getCurie().toString();
 
   try {
+    let timeout = 5000;
+    if (config.assetCount && config.assetCount > 10) {
+      timeout += (500 * (config.assetCount - 10));
```
efjacobson, line 50: @hunyoboy what does this accomplish?

efjacobson (changes requested):

src/plugins/dam/sagas/patchAssetsFlow.js
hunyoboy: @efjacobson i'm glad you mentioned. so this is based from my past update here - https://github.com/triniti/cms-js/blob/master/src/plugins/dam/sagas/reorderGalleryAssetsFlow.js#L90 w/c is @gdbrown 's idea. i'm thinking this is applicable here since both actions  are using `2pc`.  what do you think Greg?

efjacobson:
@hunyoboy when the file finishes uploading it wipes out the pending form changes https://recordit.co/rD8Mvx7j3Q

hunyoboy:
> @hunyoboy when the file finishes uploading it wipes out the pending form changes https://recordit.co/rD8Mvx7j3Q

@efjacobson yes i noticed this is an existing behaviour on test, dev and maybe prod. i mean we can address this here on this ticket if you want to. let me know.

efjacobson:
@hunyoboy if it is on prod rn then dont worry about it

src/plugins/dam/components/uploader/delegate.js
```diff
@@ -173,15 +173,21 @@ class Delegate extends AbstractDelegate {
...
       values: { credit },
     };
-    this.dispatch(initialize(this.getFormName(), { ...initialValues, credit }, 'credit'));
-    this.dispatch(patchAssets(data, files, this.config));
+    this.dispatch(initialize(
+      this.getFormName(),
+      { ...initialValues, credit: { value: credit } },
+      'credit',
+    ));
+    this.dispatch(patchAssets(data, files, config));
```
efjacobson, line 186: @hunyoboy you just want to include the assetCount, right? seems like it would be better to do something like

`this.dispatch(patchAssets(data, files, { ...this.config, assetCount: Object.keys(files).length }));`

efjacobson (changes requested):

src/plugins/dam/components/uploader/delegate.js
hunyoboy: @efjacobson nice works for me. do you really want to pass the whole config? i feel like schemas is enough currently. this is the same as the other flows ( ex. reorderGallery ). what do you think?

efjacobson: @hunyoboy i suppose it doesnt really matter, i just know the uploader is kinda tricky and was wary about replacing the config

efjacobson (approved):

PR #303, opened 2021-04-13
fix node timeout alert

aloncohenn:


src/plugins/ncr/screens/node/index.jsx
```diff
@@ -126,6 +126,8 @@ export default class AbstractNodeScreen extends React.Component {
     this.unblock();
     const { delegate } = this.props;
     delegate.componentWillUnmount();
+    clearTimeout(this.nodeAlertTimeoutId);
+    delete this.nodeAlertOpen;
```
efjacobson, line 130: @aloncohenn delete the timeout id as well

efjacobson:
closed in favor of hotfix https://github.com/triniti/cms-js/pull/304

PR #304, opened 2021-04-13
v0.13.2

efjacobson:
* Prevent "node failed to load" alert from firing after leaving node screen


hunyoboy (approved):
@rsumilang  cool stuff!

PR #305, opened 2021-04-26
A Safer Blocksmith

hunyoboy:
1. create an `OuterErrorBoundary` to wrap `blocksmith` editor as per [react doc](https://reactjs.org/docs/error-boundaries.html).
2. `stringify` inside `onError` fn has a high chance of causing a crash so added a try catch around it incase.


hunyoboy:
@efjacobson please review if you have a chance. thanks.

src/plugins/blocksmith/components/blocksmith/error-boundary/index.jsx
```diff
@@ -1,170 +1,99 @@
...
-              Component = () => <p>{message}</p>;
-          }
-          return (
-            <div className="preview-component" key={block.get('etag', block.generateEtag())}>
-              <Component />
-            </div>
-          );
-        })}
-      </div>
+      <Fallback
```
efjacobson, line 84: @hunyoboy does `Fallback` really need to be a separate component? why not just inline it in index?

src/plugins/blocksmith/components/blocksmith/error-boundary/index.jsx
```diff
@@ -1,170 +1,99 @@
...
 import { connect } from 'react-redux';
-import { ContentBlock, EditorState } from 'draft-js';
-import blockParentNode from '@triniti/cms/plugins/blocksmith/utils/blockParentNode';
-import damUrl from '@triniti/cms/plugins/dam/utils/damUrl';
+import { EditorState } from 'draft-js';
 import PropTypes from 'prop-types';
 import React from 'react';
+import blockParentNode from '@triniti/cms/plugins/blocksmith/utils/blockParentNode';
+import createDelegateFactory from '@triniti/app/createDelegateFactory';
+import delegateFactory from '@triniti/cms/plugins/blocksmith/components/blocksmith/delegate';
```
efjacobson, line 7: @hunyoboy should not use another component's delegate..why not just pass in handleStoreEditor as a prop?

efjacobson (changes requested):
@hunyoboy 

- please add an eslint exception for `import/no-named-as-default`

src/plugins/blocksmith/components/blocksmith/error-boundary/index.jsx
hunyoboy: @efjacobson good morning. it doesn't need to be but i strongly believe that it's beneficial for `ErrorBoundary` to be simple and easy to reason about as much as possible due to the reason that it is our error catcher/protector. any error that'll occur here will bubble up and can cause some issues. plus it makes unit test clearer for assertion. like oh yeah it should render a fallback if error occurs.

efjacobson: @hunyoboy it is already simple https://github.com/triniti/cms-js/blob/master/src/plugins/blocksmith/components/blocksmith/error-boundary/index.jsx#L62

having an extra component that is not reused, does not have its own selector/delegate/styles etc is just extra props drilling

src/plugins/blocksmith/components/blocksmith/error-boundary/index.jsx
hunyoboy: @efjacobson right i was just thinking that this component belongs under `blocksmith` folder so it could re-use the blocksmith's delegate. 
anyhow, since we will be passing the `handleStoreEditor` prop, we are gonna end up passing it from all the `Form` that's using the error-bounded blocksmith. Blocksmith,jsx is already using the delegate so we cannot re-use it. is that the approach you are thinking? just want to make sure.

efjacobson: @hunyoboy ahh yes, that is weird. clearest solution is to just give error-boundary its own delegate

src/plugins/raven/services/Raven.js
```diff
@@ -217,22 +217,32 @@ export default class Raven {
...
+    const errorData = args.find((arg) => arg instanceof Error) || args[0];
 
+    let error;
+    try {
+      error = JSON.stringify(errorData, Object.getOwnPropertyNames(errorData || {}));
+    } catch (e) {
+      console.error('raven::onError::error', e);
+    }
+
+    error = (error || '').replaceAll('://', '[PROTOCOL_TOKEN]').replace(/(\/\.\.)+\/?/g, '[UP_DIRECTORY_TOKEN]');
```
efjacobson, line 229: @hunyoboy dont think it is super helpful to log an empty error to the stream - what about a message like "failed stringifying errorData" or something?

src/plugins/blocksmith/components/blocksmith/error-boundary/index.jsx
```diff
@@ -1,75 +1,93 @@
...
+    // invoked first in order to have the latest/greatest prop.
+    setTimeout(() => this.handleRestoreBlocksmith(false));
     console.error(error, errorInfo);
     window.onerror(error);
   }
 
-  restoreBlocksmith(editorState, resetErrorCount = true) {
-    const { onStoreEditor: handleStoreEditor } = this.props;
-    this.setState(({ errorCount }) => ({
+  handleRestoreBlocksmith(resetErrorCount = true, excludeInvalidBlocks = true) {
```
efjacobson, line 59: @hunyoboy i think `includeInvalidBlocks` would be clearer - what is the benefit of having a negated name for a boolean?

also - im not really into changing this method to not always restore the blocksmith. it's in the name - why not put the early return logic elsewhere? kinda hard to understand as is

efjacobson (changes requested):

src/plugins/raven/services/Raven.js
hunyoboy: good morning. it won't due to L235. tested by throwing an error anywhere in blocksmith to trip `stringify`  :
 
 `const err = new Error(‘test’);
  err.me = err;
  throw err;`

or make ` stringify` return undefined.

src/plugins/blocksmith/components/blocksmith/error-boundary/index.jsx
hunyoboy: @efjacobson great eyes. i can't agree more. so because of this - https://github.com/triniti/cms-js/pull/305/commits/e90deaa3537bc0aab6e54cfdb4fcc5351a217a6c

src/plugins/blocksmith/components/blocksmith/error-boundary/index.jsx
```diff
@@ -79,92 +99,114 @@ class ErrorBoundary extends React.Component {
...
+          </kbd>
+        </CardHeader>
+        <CardBody indent>
+          <div className="blocksmith-error-boundary">
+            <p className="warning">{warning}</p>
+            <Button
+              size="md"
+              onClick={() => {
+                const result = validateBlocks(editorState);
+                this.handleRestoreBlocksmith({ ...result, editorState: result.validEditorState });
```
efjacobson, line 122: @hunyoboy since you removed

```
    if (isValid && errorCount < MAX_ERROR_COUNT) {
      this.restoreBlocksmith(validEditorState, false);
      return null;
    }
```

this would need to pass isValid in, eh?

that being said - is all this change really necessary? i think just wrapping the entire component in the error boundary (which is a great idea btw) and the Raven changes is sufficient

efjacobson (changes requested):

src/plugins/blocksmith/components/blocksmith/error-boundary/index.jsx
hunyoboy: @efjacobson it'll still work but the main reason why i'm modifying it here since i believe it's not advisable or we should avoid if possible to put any side effects inside `render` function. also react [doc](https://reactjs.org/docs/react-component.html#render) here states clearly: 

> The render() function should be pure, meaning that it does not modify component state, it returns the same result each time it’s invoked, and it does not directly interact with the browser.

If you need to interact with the browser, perform your work in componentDidMount() or the other lifecycle methods instead. Keeping render() pure makes components easier to think about.

and on their error boundary [demo](https://codepen.io/gaearon/pen/wqvxGa?editors=0010). 

also there's a good discussion [here](https://stackoverflow.com/questions/48226268/calling-setstate-in-react-from-render-method/52024407#:~:text=Calling%20setState()%20inside%20render,state%20and%2For%20props%20directly) about it.

efjacobson: @hunyoboy why does `handleRestoreBlocksmith` even have the `isValid` parameter? it is never provided

this approach seems less clear, im trying to understand the real benefit. i really think just moving the error-boundary and the Raven change is the win here

efjacobson (changes requested):

src/plugins/blocksmith/components/blocksmith/error-boundary/index.jsx
efjacobson: @hunyoboy why does `handleRestoreBlocksmith` even have the `isValid` parameter? it is never provided

this approach seems less clear, im trying to understand the real benefit. i really think just moving the error-boundary and the Raven change is the win here

efjacobson: @hunyoboy why does `handleRestoreBlocksmith` even have the `isValid` parameter? it is never provided

this approach seems less clear, im trying to understand the real benefit. i really think just moving the error-boundary and the Raven change is the win here

efjacobson: @hunyoboy also, the way it is now is not going to create an infinite loop as discussed in that stackoverflow thread

hunyoboy: > @hunyoboy why does `handleRestoreBlocksmith` even have the `isValid` parameter? it is never provided
> 

@efjacobson yes it is because we are now setting that within the state. and yes we need to pass it now and it's kind of a drawback but not really since we're just computing this exactly as how many times we needed it regardless how many times the component is re-rerendered. current implementation is we're pulling it from `validateBlocks` w/c is inside render. and this render can be called many, many times. if you intentionally throw an exception and render an error-boundary, you will see that the component will still be re-rendered multiple times and will unnecessarily invoke 'validateBlocks` ( ex. raven can re-render component ). i don't think you'll agree with that. whereas if we extract it outside render, we do  not make those unnecessary calls.

efjacobson: @hunyoboy 

> yes it is because we are now setting that within the state. and yes we need to pass it now and it's kind of a drawback but not really since we're just computing this exactly as how many times we needed it regardless how many times the component is re-rerendered
where is `isValid` passed into `handleRestoreBlocksmith`?

> if you intentionally throw an exception and render an error-boundary, you will see that the component will still be re-rendered multiple times and will unnecessarily invoke 'validateBlocks`
does it matter? let me ask you this - what problem are you trying to solve? is the additional complexity worth it to prevent extra renders in this edge case scenario? now this component has a lot more state that is just unnecessary - it can be derived from props..it is duplicate state and that should be avoided at all costs

src/plugins/blocksmith/components/blocksmith/index.jsx
```diff
@@ -1,1862 +1,10 @@
...
-          )}
-        </CardBody>
-      </Card>
-    );
-  }
-}
-
-export default connect(selector, createDelegateFactory(delegateFactory))(Blocksmith);
+export default ({ ...rest }) => (
+  <OuterErrorBoundary>
```
efjacobson, line 7: @hunyoboy can you inline this? odd to have `OuterErrorBoundary` inside the `error-boundary` dir

src/plugins/blocksmith/components/blocksmith/index.jsx
```diff
@@ -1,1862 +1,12 @@
...
-          </>
-          )}
-        </CardBody>
-      </Card>
-    );
-  }
-}
-
-export default connect(selector, createDelegateFactory(delegateFactory))(Blocksmith);
+export default ({ ...rest }) => (
```
efjacobson, line 6: @hunyoboy spread into object is kind of halfway between two options:

```
export default (props) => (
  <OuterErrorBoundary
    formName={props.formName}
  >
    <Blocksmith {...props} />
  </OuterErrorBoundary>
);
```

or

```
export default ({ formName, ...rest }) => (
  <OuterErrorBoundary
    formName={formName}
  >
    <Blocksmith formName={formName} {...props} />
  </OuterErrorBoundary>
);
```

efjacobson (changes requested):
looks good aside from my small note. we should set up a demo with Sam for this, given its importance

efjacobson (approved):

PR #314, opened 2021-05-14
Add eme-form-block to Blocksmith

albertgcsula:
- A user should be able to enter an eme form ref as input.
- EME form ref must match pattern: `vendor:label:id`, eg. `eme:solicit:103bbf32-f3b4-415e-a85c-0735db7ca73d`
- Form block data should resemble:
`{
      "_schema": "pbj:thereal:canvas:block:eme-form-block:1-0-0",
      "etag": "b7a8a76960e13d77e7d6d9d9b6faa2c3",
      "aside": false,
      "form_ref": "eme:solicit:103bbf32-f3b4-415e-a85c-0735db7ca63d"
    }`


src/plugins/blocksmith/components/eme-form-block-modal/index.jsx
```diff
@@ -0,0 +1,141 @@
...
+
+  handleEditBlock() {
+    const { onEditBlock, toggle } = this.props;
+    onEditBlock(this.setBlock());
+    toggle();
+  }
+
+  handleChangeFormRef(event) {
+    const input = event.target.value;
+    const id = getEmeFormBlockRef(input);
```
efjacobson, line 67: @albertgcsula `id` is an odd name for this variable eh? its null or a noderef string. and the util itself - why not just have an `isValidNodeRef` and use `NodeRef.fromString` inside?

efjacobson (changes requested):

src/plugins/blocksmith/components/eme-form-block-modal/index.jsx
albertgcsula: @efjacobson code updated.

efjacobson (approved):

PR #316, opened 2021-05-22
SEO desc char

rsumilang:


gdbrown:
@rsumilang imo this is something that should be just just on the field itself, like in seo field or make a custom field just for seo description.  this needs to be a lot more isolated.

rsumilang:
@gdbrown I took inspiration here from how the news title field was implemented. Should that have it's own field for title as well?

gdbrown:
@rsumilang i figured so, but this strategy needs to be rethought.  too much code for such a small feature/benefit.  a single redux field can do this on its own without any other code needing to change.

PR #319, opened 2021-05-25
SEO desc lengths

rsumilang:
Adjust SEO Description Character Counter To Turn Yellow At 140 Characters and Red at 160.


src/plugins/common/components/seo-fields/index.jsx
```diff
@@ -7,8 +7,8 @@ import Message from '@gdbots/pbj/Message';
 import PropTypes from 'prop-types';
 import React, { useState } from 'react';
 import SelectField from '@triniti/cms/components/select-field';
-import TextareaField from '@triniti/cms/components/textarea-field';
 import TextField from '@triniti/cms/components/text-field';
+import MetaDescriptionTextareaField from './meta-description-textarea-field';
```
gdbrown, line 11: @rsumilang call it `MetaDescriptionField` (filename as well)

gdbrown (changes requested):
@rsumilang much simpler!  one minor change

rsumilang:
@gdbrown cool

gdbrown (changes requested):
@rsumilang change the file to MetaDescriptionField.jsx, we only name the dirs with slug-case, files within are PascalCase or camelCase

rsumilang:
done

gdbrown (approved):

PR #322, opened 2021-05-28
paste from local storage

bhsiao55:
* Allow users to copy from one tab and paste in another.


src/plugins/blocksmith/components/blocksmith/Blocksmith.jsx
```diff
@@ -1356,11 +1372,36 @@ class Blocksmith extends React.Component {
   }
 
   /**
-   * If there is a copied block available in redux, use it to create a draft block with it as the
-   * data payload.
+   * Dereferences any nodes in the copied block that are referenced
+   */
+  derefCopiedBlockNodes() {
```
efjacobson, line 1377: @bhsiao55 cant you do this only on paste? or, even better, put into the placeholder components themselves so blocksmith doesnt need more overhead

efjacobson (changes requested):

src/plugins/blocksmith/components/blocksmith/Blocksmith.jsx
bhsiao55: @efjacobson I tried calling derefCopiedBlockNodes on paste but for example when pasting an article block the title and image would render at different times so to eliminate that effect I called it in componentDidUpdate.  If I placed it in the individual placeholder components I suspect it will have that same effect.

efjacobson: @bhsiao55 its only a very short time though eh? why not just not render the placeholder until the title/image are available?

bhsiao55: @efjacobson yea it's for a short time.  ok I will move the loading of the nodes to the preview component

bhsiao55:
@efjacobson moved code to place holder components.  please review again when you have time thanks.

efjacobson (approved):

bhsiao55:
@gdbrown can you merge and tag if it looks good to you? thanks

efjacobson:
@bhsiao55 can you get the stuff from release-next in here first?

bhsiao55:
@efjacobson release-next is in here now

PR #331, opened 2021-06-23
v0.14.1

gdbrown:
* Add `layout` field to gallery screen.
* Sort users by first name asc on search users screen.
* Update viewportMaxHeight css variable for DAM uploader.
* Notify user when session expired and give instructions on how to prevent losing work.


efjacobson (dismissed):
nice

PR #333, opened 2021-06-25
v0.15.0

gdbrown:
* Remove pbjx timeout checks on sagas as they are no longer needed. (note, sagas disappear completely in next major rev for 99% of operations).
  The server runs write operations sync instead of async so any error will be reported immediately and we don't need to wait for the raven events to determine success.


src/plugins/ncr/selectors/getNode.js
```diff
@@ -16,7 +16,7 @@ export default (state, ref) => {
 
   let nodeRef = ref;
   if (!(nodeRef instanceof NodeRef)) {
-    nodeRef = NodeRef.fromString(ref);
+    nodeRef = NodeRef.fromString(`${ref}`);
```
efjacobson, line 19: nice

efjacobson (approved):

PR #334, opened 2021-07-01
v0.16.0

aloncohenn:
* Add support for auto posting to twitter


src/plugins/iam/components/twitter-app-fields/index.jsx
```diff
@@ -0,0 +1,34 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Field } from 'redux-form';
+import { Card, CardBody, CardHeader } from '@triniti/admin-ui-plugin/components';
+import TextField from '@triniti/cms/components/text-field';
+import TextAreaField from '@triniti/cms/components/textarea-field';
+
+const TwitterAppFields = ({ isEditMode }) => [
+  <Card key="basic">
```
efjacobson, line 9: @aloncohenn nit: i generally prefer using fragments over keyed array entries these days, you see keys around since fragments werent fully supported when we started this project

src/plugins/iam/services/TwitterAppSubscriber.js
```diff
@@ -0,0 +1,136 @@
...
+  /**
+   * Occurs before validation and is used to add warnings.
+   *
+   * @param {FormEvent} formEvent
+   */
+  onWarnForm(formEvent) {
+    const data = formEvent.getData();
+    const { oauthConsumerSecret, oauthTokenSecret } = data;
+
+    // TODO: fix warning, it's not working here or in apple news subscriber
```
efjacobson, line 48: @aloncohenn what is the warning?

src/plugins/iam/services/TwitterAppSubscriber.js
```diff
@@ -0,0 +1,136 @@
...
+   */
+  onSubmitForm(formEvent) {
+    const data = formEvent.getData();
+    const node = formEvent.getMessage();
+
+    if (node.isFrozen()) {
+      return;
+    }
+
+    if (!formEvent.getProps().isCreateForm) {
```
efjacobson, line 112: @aloncohenn bit cleaner to do an early return here

efjacobson (changes requested):

src/plugins/iam/components/twitter-app-fields/index.jsx
aloncohenn: fixed

src/plugins/iam/services/TwitterAppSubscriber.js
aloncohenn: removed comment this was fixed

src/plugins/iam/services/TwitterAppSubscriber.js
aloncohenn: @efjacobson this is consistent with how all the other ones do it, do we want to make this one different or keep the same?

efjacobson: @aloncohenn hmm..dont think there is a reason they have to be like that. anyways not worth it, leave as is

efjacobson:
@aloncohenn if you want this as a hotfix the pr title/desc should update

efjacobson (approved):

PR #336, opened 2021-08-05
Add sports theme

tbone1020:


tbone1020:
Theme field on gallery pages are not needed right now.

PR #339, opened 2021-08-30
Fix Keyboard Paste on YouTube Video Block

albertgcsula:
- Fix ctr + v command in YouTube video block modal.


albertgcsula:
branch seems borked.

PR #342, opened 2021-09-20
Code clone autocheck

tbone1020:
https://github.com/wb-apps/tmz-cms/issues/1326


src/plugins/curator/sagas/cloneNodeFlow.js
```diff
@@ -90,6 +88,11 @@ function* clone({ history, node }) {
       }
     }
   });
+
+  if (clonedNode.schema().className.indexOf('CodeWidget') !== -1) {
```
efjacobson, line 92: @tbone1020 check for the mixin like you see in other places in this file

gdbrown (commented):
@tbone1020 the clonenode flow is a generic feature, it can't really have site specific rules or schema specific rules.  if you want ot have all "new" code widgets have a certain field set a certain way you can create a site specific event listener in the enrich phase. e.g.

src/plugins/ncr/sagas/index.js
```diff
@@ -2,7 +2,7 @@ import { all, fork, takeLatest } from 'redux-saga/effects';
 import { actionTypes } from '../constants';
 
 import batchOperationFlow from './batchOperationFlow';
-import cloneNodeFlow from './cloneNodeFlow';
+import cloneNodeFlow from '../../curator/sagas/cloneNodeFlow';
```
efjacobson, line 5: @tbone1020 fully move this over to curator, so instead of importing like this (here in ncr) you would import it in curator/sagas. all of the related code should move too like the action and whatnot

efjacobson (changes requested):

PR #349, opened 2022-01-04
Bump nvmrc & pin lodash/lodash-es

efjacobson:


gdbrown (approved):

gdbrown:
@efjacobson not sure the problem with esm exists if you don't upgrade node version.  btw, in new versions of schemas and pbjx i've eliminated the lodash and just use lodash-es direct.  that's coming at some point and also removed the babel plugins for lodash

efjacobson:
@gdbrown yeah i think with the current version it does not have the problem, but now that some folks are on npm 8.3.0 it is complaining about anything less than v12. tried v18 and v16 first and they introduced all kinds of other problems

gdbrown:
@efjacobson indeed, ESM updates are total pain in node and babel, testing, etc.  i've had to do some real weird stuff to get it work, like using dynamic imports to get tape tests to work.  e.g. https://github.com/gdbots/pbjx-js/blob/master/tests/bootstrap.cjs and using file package reference instead of babel module resolver since that doesn't work with esm and cjs mixtures.  https://github.com/triniti/schemas/blob/master/package.json#L29

upgrades are mostly pain.

PR #351, opened 2022-01-12
Customizable bs sidebar

rsumilang:
These updates to the blocksmith sidebar allow for further customization of the button groups.

- Button groups may be placed anywhere (eg: above vendor group).
- Buttons can be listed in any order.
- Vendor button group can be deleted.
- Other group can be turned off.

In addition, the search was modified a tad. Buttons can be placed in multiple groups thus have duplicates. This is nice for the default sidebar view but no need to have duplicates when searching. The search results will list all buttons alphabetically and dedupe them.

Note: This is a breaking change as previous sidebar configs are no longer compatible.

### DIFF
--- a/src/plugins/blocksmith/components/sidebar/config.js
+++ b/src/plugins/blocksmith/components/sidebar/config.js
@@ -1,8 +1,9 @@
+/* globals APP_VENDOR */
 export default {
   enableOther: true,
   groups: [
     {
-      header: 'Triniti',
+      header: APP_VENDOR,
       buttons: [
         'article-block',
         'audio-block',
--- a/src/plugins/blocksmith/components/sidebar/config.js
+++ b/src/plugins/blocksmith/components/sidebar/config.js
@@ -1,42 +1,32 @@
-const vendorButtonTypes = [
-  'article-block',
-  'audio-block',
-  'divider-block',
-  'document-block',
-  'gallery-block',
-  'heading-block',
-  'image-block',
-  'page-break-block',
-  'poll-block',
-  'poll-grid-block',
-  'quote-block',
-  'text-block',
-  'video-block',
-];
-
-export { vendorButtonTypes };
-
-/**
- * Config for how the sidebar buttons are displayed/organized/made searchable
- *
- *   - header will be a string in the PopoverHeader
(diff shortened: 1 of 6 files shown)


efjacobson:
@rsumilang can you make this not a breaking change by continuing to export default but also exporting a new config?

efjacobson:
@rsumilang also - is it actually a breaking change? this would be the first time a vendor is customizing the config

efjacobson:
@rsumilang but none of the vendors are currently using a custom config - when they bring in this new triniti version they will automatically update and the scottish vendor will be only one with a resolved import

rsumilang:
@efjacobson ah right on. if no vendors are using the config then it should be a seamless upgrade :) how does the code look otherwise then?

src/plugins/blocksmith/components/sidebar/config.js
```diff
@@ -1,42 +1,32 @@
...
-    matchRegEx: new RegExp(`(twitter|facebook|instagram|^(${vendorButtonTypes.join('|')}))`),
-    doesMatch: false,
-    replaceRegEx: new RegExp('block', 'g'),
-  },
-];
+export default {
+  enableOther: true,
+  groups: [
+    {
+      header: 'Triniti',
```
efjacobson, line 5: @rsumilang i think this is going to force each vendor to have its own config? you're going to want to use the localizationMap here https://github.com/triniti/cms-js/blob/master/demo/src/config/localizationMap.js

src/plugins/blocksmith/components/sidebar/config.js
```diff
@@ -1,8 +1,9 @@
+/* globals APP_VENDOR */
 export default {
   enableOther: true,
   groups: [
     {
-      header: 'Triniti',
+      header: APP_VENDOR,
```
efjacobson, line 6: @rsumilang need to use the localizationmap here

rsumilang: Localization map can be updated on a per vendor app basis.

efjacobson (commented):
@rsumilang not sure this needs to be a release right now, just go into release-next

efjacobson (approved):

PR #357, opened 2022-03-02
Added top level className prop to DateFieldPicker

digitalartcrew:
Added top level className prop for handing the styling of `DateFieldPicker`


src/components/date-picker-field/README.md
```diff
@@ -11,6 +11,10 @@ This prop is provided by redux-form, a name prop must be passed to `Field` compo
 
 ###Optional Props
 
+**className : String**
+
+Default value - ''
```
efjacobson, line 16: @digitalartcrew the default prop is not assigned in the component

efjacobson (changes requested):
@digitalartcrew if you want to go straight into master you'll need to update package.json and package-lock.json, changelog, etc. ref one of the previous release PRs

src/plugins/common/components/seo-fields/index.jsx
```diff
@@ -32,7 +32,7 @@ const SeoFields = ({ areLinkedImagesAllowed, isEditMode, node }) => {
     }
   };
   return (
-    <Card>
+    <Card className="seo-fields">
```
efjacobson, line 35: @digitalartcrew do you need this one as well as the other?

digitalartcrew: The purpose of this class was to replicate the BEM css as in other components and it's a generic class to simply identify the component: https://github.com/triniti/cms-js/blob/master/src/components/key-values-field/index.jsx

The other class is so when the component is imported, we only update the css.

efjacobson: @digitalartcrew in that case i would name the other class `seo-fields__seo-published-at` since that is the actual name of the field in the schema

efjacobson:
@digitalartcrew looks good to me, what say you @gdbrown ?

efjacobson:
@gdbrown had a chance to look at this?

PR #358, opened 2022-03-29
v0.18.2

digitalartcrew:
* Fix flickering block previews

### DIFF
--- a/src/plugins/blocksmith/components/facebook-post-block-modal/index.jsx
+++ b/src/plugins/blocksmith/components/facebook-post-block-modal/index.jsx
@@ -56,22 +56,23 @@ export default class FacebookPostBlockModal extends React.Component {
     this.handleChangeTime = this.handleChangeTime.bind(this);
     this.handleEditBlock = this.handleEditBlock.bind(this);
     this.handleChangeAside = this.handleChangeAside.bind(this);
+    this.handleMouseOut = this.handleMouseOut(this);
   }
 
-  handleMouseOut = e => {
+  handleMouseOut(e) {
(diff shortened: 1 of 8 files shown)


src/plugins/blocksmith/components/facebook-post-block-modal/index.jsx
```diff
@@ -58,10 +58,20 @@ export default class FacebookPostBlockModal extends React.Component {
...
+  handleMouseOut = e => {
+    e.stopPropagation();
+  };
+
   componentDidMount() {
     setTimeout((t) => {
       t.inputElement.focus();
     }, 0, this);
+    
+    window.addEventListener('mouseout',(e) => this.handleMouseOut(e), true);
```
efjacobson, line 70: @digitalartcrew this is still an anonymous function.. do something like 

`window.addEventListener('mouseout', this.handleMouseOut, true);`

src/plugins/blocksmith/components/pinterest-pin-block-modal/index.jsx
```diff
@@ -63,10 +63,20 @@ export default class PinterestPinBlockModal extends Component {
     this.handleChangeSelect = this.handleChangeSelect.bind(this);
   }
 
+  handleMouseOut = e => {
```
efjacobson, line 66: @digitalartcrew use the same syntax for this method as the other class methods eg

```
  handleMouseOut(e) {
    e.stopPropagation();
  }
```

and then youll use it as this.handleMouseOut

efjacobson:
looking good, over to you @gdbrown

PR #360, opened 2022-06-18
v0.18.4

digitalartcrew:


src/plugins/dam/components/uploader/Uploader.jsx
```diff
@@ -92,6 +92,24 @@ class Uploader extends React.Component {
...
+
+    const isPropsEqual = (JSON.stringify(currentValues) === JSON.stringify(nextProps.currentValues));
+
+    if (!isPropsEqual) {
+      return false;
+    }
+
+    const hasUpdatedImage = activeHashName !== nextProps.activeHashName;
+
+    if (hasUpdatedImage || hasFilesProcessing) {
```
efjacobson, line 106: @digitalartcrew could just return `!!(hasUpdatedImage || hasFilesProcessing)`

src/plugins/dam/components/uploader/Uploader.jsx
```diff
@@ -252,6 +265,57 @@ class Uploader extends React.Component {
...
+
+    this.state = {
+      isDisabled: true
+    }
+  }
+
+  toggleDisable(e) {
+    setTimeout(() => {
+      const { delegate } = this.props;
+      const hasCredits = !!document.querySelector('.select__single-value');
```
efjacobson, line 282: @digitalartcrew dom querying in a react app is an anti pattern only to be used as a last resort - there must be another piece of state or prop that indicates the same thing

src/plugins/dam/components/uploader/Uploader.jsx
```diff
@@ -252,6 +265,57 @@ class Uploader extends React.Component {
...
+      const { description, displayTitle, altText, credit } = delegate.component.props.currentValues
```
efjacobson, line 283: @digitalartcrew why not just pass it down to the button?

src/plugins/dam/components/uploader/Uploader.jsx
```diff
@@ -252,6 +265,57 @@ class Uploader extends React.Component {
   }
 }
 
+class SaveButton extends React.Component {
```
efjacobson, line 268: @digitalartcrew why does this need to be a separate component?

efjacobson (changes requested):
@digitalartcrew added some comments. ill try to take a closer look at this later to see if there is a more elegant way

src/plugins/dam/components/uploader/Uploader.jsx
digitalartcrew: This was so the button could have it's own state that wouldn't re-render the entire component

src/plugins/dam/components/uploader/Uploader.jsx
```diff
@@ -92,6 +92,24 @@ class Uploader extends React.Component {
     this.handleToggleUploader = this.handleToggleUploader.bind(this);
   }
 
+  shouldComponentUpdate(nextProps){
+    const { currentValues, hasFilesProcessing, activeHashName } = this.props;
+
+    const isPropsEqual = (JSON.stringify(currentValues) === JSON.stringify(nextProps.currentValues));
+
+    if (!isPropsEqual) {
+      return false;
```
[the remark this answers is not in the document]
digitalartcrew: Correct, the update is causing the re-render. So I only capture the values when the form is actually saved. Values have been verified: https://recordit.co/IYx2PEsokL

src/plugins/dam/components/uploader/Uploader.jsx
digitalartcrew: https://recordit.co/b4lpOFGKts

src/plugins/dam/components/uploader/Uploader.jsx
digitalartcrew: Yes this was interesting because the was not an actual input or select element to capture value

src/plugins/dam/components/uploader/FileList.jsx
```diff
@@ -35,19 +35,7 @@ class FileList extends React.Component {
...
-    if (document.documentElement.clientHeight > 792) {
-      this.ensureActiveItemVisible();
-    }
-  }
-
-  ensureActiveItemVisible() {
-    if (this.activeFileItem) {
-      scrollIntoViewIfNeeded(this.activeFileItem, { duration: 500 });
-    }
+    setTimeout(() => this.ensureActiveItemVisible(), 2000);
```
efjacobson, line 38: @digitalartcrew this wont stop it from running on every update, it just delays it. use lodash throttle to make it so that it only fires at most once every 5 seconds https://lodash.com/docs/4.17.15#throttle

src/plugins/dam/components/uploader/FileList.jsx
```diff
@@ -65,6 +53,28 @@ class FileList extends React.Component {
     }
   }
 
+  ensureActiveItemVisible() {
+    if (!this.activeFileItem) {
+      return;
+    }
+    computeScrollIntoView(this.activeFileItem, { scrollMode: 'if-needed' }).forEach((action) => {
+      if (action.el.className !== 'dam-file-list') { // fixme: use the class name from a ref instead of hardcoding
```
efjacobson, line 61: @digitalartcrew do the todo...make this a ref instead of hardcoding the class name

efjacobson (changes requested):

src/plugins/dam/components/uploader/FileList.jsx
```diff
@@ -65,6 +55,34 @@ class FileList extends React.Component {
...
+        el.scroll({
+          top,
+          behavior: 'smooth',
+        });
+      } else {
+        el.scrollTop = top;
+      }
+    });
+
+    throttle(() => throttledComputeScrollIntoView(), 5000);
```
efjacobson, line 83: @digitalartcrew 
- you dont have to "double function" this. `() => someFunction()` is the same thing as `someFunction`. the end result is the same
- doesnt look like `throttledComputeScrollIntoView` exists

src/plugins/dam/components/uploader/FileList.jsx
```diff
@@ -32,22 +33,11 @@ class FileList extends React.Component {
...
-    if (document.documentElement.clientHeight > 792) {
-      this.ensureActiveItemVisible();
-    }
-  }
-
-  ensureActiveItemVisible() {
-    if (this.activeFileItem) {
-      scrollIntoViewIfNeeded(this.activeFileItem, { duration: 500 });
-    }
+   this.ensureActiveItemVisible();
```
efjacobson, line 40: @digitalartcrew this is where the throttle should go, ref some other examples in teh codebase like https://github.com/triniti/cms-js/blob/656d22fc77c88d6f3269e6403048b8856847c433/src/plugins/iam/services/Authenticator.js#L275

src/plugins/dam/components/uploader/FileList.jsx
```diff
@@ -65,6 +55,34 @@ class FileList extends React.Component {
     }
   }
 
+  ensureActiveItemVisible() {
+    if (!this.activeFileItem) {
+      return
+    }
+
+    const currentClassName = this.damFileList.current.className;
```
efjacobson, line 63: @digitalartcrew dont think you really need a separate variable here, you can just put it into the if in the loop. also - is `this.damFileList` ever not set yet by the time we get to this comparison? i know that `this.activeFileItem` is not ready the first time

src/plugins/dam/components/uploader/FileList.jsx
```diff
@@ -32,22 +33,11 @@ class FileList extends React.Component {
     super(props);
 
     this.handleOnChange = this.handleOnChange.bind(this);
+    this.damFileList = React.createRef();
```
efjacobson, line 36: @digitalartcrew id probably just call this fileList, the class name has dam in it because css is global and whatnot

src/plugins/dam/components/uploader/FileList.jsx
```diff
@@ -65,6 +55,34 @@ class FileList extends React.Component {
...
 
+  ensureActiveItemVisible() {
+    if (!this.activeFileItem) {
+      return
+    }
+
+    const currentClassName = this.damFileList.current.className;
+
+    computeScrollIntoView(this.activeFileItem, { scrollMode: 'if-needed' }).forEach((action) => {
+      if (action.el.className !== currentClassName) {
```
efjacobson, line 66: @digitalartcrew comment here would be good for those who come after us, something like "computeScrollIntoView thinks we need to scroll more elements than is actually necessary/desired"

efjacobson (changes requested):
@digitalartcrew need to add changelog entry in pr body

src/plugins/dam/components/uploader/FileList.jsx
```diff
@@ -32,22 +33,21 @@ class FileList extends React.Component {
     super(props);
 
     this.handleOnChange = this.handleOnChange.bind(this);
+    this.fileList = React.createRef();
+    this.ensureActiveItemVisible = this.ensureActiveItemVisible.bind(this);
```
efjacobson, line 37: @digitalartcrew you can just throttle it here in the constructor, check out the example here https://medium.com/@markotroskot/debouncing-and-throttling-in-react-feb84d84bf28

src/plugins/dam/components/uploader/FileList.jsx
```diff
@@ -32,22 +33,21 @@ class FileList extends React.Component {
...
-    // or less and prevent the scrolling into view from happening.
-    if (document.documentElement.clientHeight > 792) {
-      this.ensureActiveItemVisible();
-    }
+    const throttleActiveItemVisibility = throttle(this.ensureActiveItemVisible, 5000);
+    throttleActiveItemVisibility();
+
+    // fix -- description text is reset while files are processing
+
+    ['input', 'click'].forEach(event => document.addEventListener(event,throttleActiveItemVisibility));
```
efjacobson, line 46: @digitalartcrew
- why do you need to check on these events?
- if you do end up needing it you would only need to add it once, not on every update

src/plugins/dam/components/uploader/FileList.jsx
```diff
@@ -32,22 +33,21 @@ class FileList extends React.Component {
...
   }
 
-  ensureActiveItemVisible() {
-    if (this.activeFileItem) {
-      scrollIntoViewIfNeeded(this.activeFileItem, { duration: 500 });
-    }
+  componentWillUnmount(){
+    ['input', 'click'].forEach(event =>  document.removeEventListener(event, this.ensureActiveItemVisible));
   }
```
efjacobson, line 51: @digitalartcrew this is not the same function that you added..

src/plugins/dam/components/uploader/FileList.jsx
```diff
@@ -65,6 +65,34 @@ class FileList extends React.Component {
...
+    
+    const currentClassName = !!this.fileList.current.className ? this.fileList.current.className : false;
+
+    if(!currentClassName){
+      return
+    }
+
+    // computeScrollIntoView thinks we need to scroll more elements than is actually necessary/desired
+    computeScrollIntoView(this.activeFileItem, { scrollMode: 'if-needed' }).forEach(({el, top }) => {
+      if (el.className !== currentClassName) {
```
efjacobson, line 81: @digitalartcrew just do

`if (el.className !== this.fileList.current.className) {`

src/plugins/dam/components/uploader/FileList.jsx
```diff
@@ -65,6 +65,34 @@ class FileList extends React.Component {
...
+    }
+    
+    const currentClassName = !!this.fileList.current.className ? this.fileList.current.className : false;
+
+    if(!currentClassName){
+      return
+    }
+
+    // computeScrollIntoView thinks we need to scroll more elements than is actually necessary/desired
+    computeScrollIntoView(this.activeFileItem, { scrollMode: 'if-needed' }).forEach(({el, top }) => {
```
efjacobson, line 80: @digitalartcrew I believe that destructuring el here causes a lint warning

efjacobson (changes requested):

efjacobson:
looks good, over to you @gdbrown

gdbrown (approved):

PR #362, opened 2022-08-03
fix toolbar visibility issue

meghedi:
https://github.com/tmz-apps/tmz-cms/issues/1373


src/plugins/blocksmith/components/blocksmith/Blocksmith.jsx
```diff
@@ -1197,6 +1197,7 @@ class Blocksmith extends React.Component {
     this.setState(() => ({
       modalComponent,
     }));
+    document.querySelector('.inline-toolbar').style.visibility = 'hidden';
```
efjacobson, line 1200: @meghedi do it the react way with conditional rendering - just dont include the toolbar at all in the dom when the modal is open

src/plugins/blocksmith/components/blocksmith/Blocksmith.jsx
```diff
@@ -1688,6 +1689,7 @@ class Blocksmith extends React.Component {
       blockButtonsStyle,
       editorState,
       errors,
+      isDirty,
```
efjacobson, line 1692: @meghedi do you need this?

efjacobson (changes requested):

src/plugins/blocksmith/components/blocksmith/Blocksmith.jsx
meghedi: Done.

src/plugins/blocksmith/components/blocksmith/Blocksmith.jsx
meghedi: @efjacobson I tried using react conditions like the change that I commented out, but still as soon as the modal is closed the toolbar shows up on the bottom. but with styling it does not. still working to find a way, I can show you when you have time.

efjacobson: @meghedi looks like there is an isVisible prop that it can take, try that https://github.com/draft-js-plugins/draft-js-plugins/blob/master/packages/inline-toolbar/src/components/Toolbar/index.tsx#L32

demo/package-lock.json
```diff
@@ -1,8283 +1,8 @@
 {
   "name": "demo",
   "version": "0.1.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
efjacobson, line 4: @meghedi dont revert the lockfileVersion

package-lock.json
```diff
@@ -1,12526 +1,8 @@
 {
   "name": "@triniti/cms",
   "version": "0.18.4",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
efjacobson, line 4: @meghedi no revert

src/plugins/blocksmith/components/blocksmith/Blocksmith.jsx
```diff
@@ -1258,25 +1259,29 @@ class Blocksmith extends React.Component {
...
-          isFreshBlock={isFreshBlock}
-          isOpen
-          node={node}
-          onAddBlock={this.handleAddCanvasBlock}
-          onEditBlock={this.handleEditCanvasBlock}
-          toggle={this.handleCloseModal}
-        />
-      ));
+      this.setState(() => ({
+        editorState: collapseSelection(editorState),
```
efjacobson, line 1269: @meghedi add a comment here about why we are doing this

efjacobson (changes requested):

demo/package-lock.json
meghedi: Done.

package-lock.json
meghedi: Done.

src/plugins/blocksmith/components/blocksmith/Blocksmith.jsx
meghedi: Done.

efjacobson (approved):

PR #364, opened 2023-01-13
v0.18.5

meghedi:
* When Updating a Block in CMS, the Toolbar Becomes Sticky At Bottom of Story.


package-lock.json
```diff
@@ -1,12526 +1,8 @@
 {
   "name": "@triniti/cms",
-  "version": "0.18.4",
-  "lockfileVersion": 2,
+  "version": "0.18.5",
+  "lockfileVersion": 1,
```
efjacobson, line 4: @meghedi dont want to go backwards on the version

meghedi: @efjacobson @gdbrown  updated, now it uses `"lockfileversion": 3

efjacobson (approved):
looks good, @gdbrown this is pretty small but note the lockfile version has been incremented

PR #368, opened 2023-06-09
## v0.18.8

digitalartcrew:
* remove ability to add updated_date to blocks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 
 ## v0.18.8
-* Added selector to hide date picker field
+* remove ability to add updated_date to blocks
 
 
 ## v0.18.7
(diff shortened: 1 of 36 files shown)


rsumilang:
@digitalartcrew I haven't forgotten about this. Waiting for a final sign off before proceeding.

rsumilang:
@digitalartcrew There are some loose ends hanging around in the source still. For example, the CustomizeOptions for Article Block Modal still requires `hasUpdatedDate` though it no longer makes use of it. Then, you need to double check with each index file that uses CustomizeOptions to remove those props and associated state variables there that are no longer needed.

efjacobson:
@digitalartcrew @rsumilang hold off on this until we hear back from acme

PR #369, opened 2023-06-27
v0.18.8

digitalartcrew:
* Most Content Blocks are not being added in the correct places - Version 3


src/plugins/blocksmith/components/blocksmith/Blocksmith.jsx
efjacobson, line 1: @digitalartcrew add another outcome to handleAddCanvasBlock such that if the activeBlockKey is null, insert the incoming block at the end. that way at least it gets added and they can reorder instead of having to start over with a "new" block

efjacobson:
@digitalartcrew also since this is a release PR it needs to follow the same convention as always for the PR body (mirror the changelog)

efjacobson (approved):

efjacobson:
@gdbrown over to you once again

PR #371, opened 2024-03-11
V2.x meg

meghedi:


src/plugins/ovp/components/livestreams-screen/LivestreamsCard.jsx
```diff
@@ -108,23 +108,33 @@ const LivestreamsCard = ({ nodes, metas, reloadChannelState }) => nodes.map((nod
...
-      await delay(2000);
-      dispatch(sendAlert({
-       type: 'success',
-       isDismissible: true,
-       delay: 3000,
-       message: 'Success! The MediaLive Channel was started.',
-      }));
-    }catch(error){
-      await progressIndicator.close();
+      const worker = app.getParameter('raven.worker');
```
gdbrown, line 111: @meghedi you can subscribe to the pbjx event listener rather than going through worker.  the worker (should be) sending the pbjx events to the app.  we want to minimize any direct listening to workers outside of raven, which is the service handling that interaction.

efjacobson:
@meghedi @rsumilang you'll want to have the listeners remove themselves from the dispatcher when they fire

src/plugins/ovp/components/livestreams-screen/LivestreamsCard.jsx
```diff
@@ -108,23 +110,31 @@ const LivestreamsCard = ({ nodes, metas, reloadChannelState }) => nodes.map((nod
...
-       message: 'Success! The MediaLive Channel was started.',
-      }));
-    }catch(error){
-      await progressIndicator.close();
+
+      const eventName = 'triniti:ovp.medialive:event:channel-started.raven';
+      const dispatcher = app.getDispatcher();
+      const cleanup = () => {
+        progressIndicator.close();
+        delay(2000);
```
efjacobson, line 118: @meghedi this delay is not necessary anymore, right? especially since it is not even being awaited

src/plugins/ovp/components/livestreams-screen/LivestreamsCard.jsx
```diff
@@ -43,6 +44,7 @@ const LivestreamsCard = ({ nodes, metas, reloadChannelState }) => nodes.map((nod
   const canViewIngests = policy.isGranted(`${APP_VENDOR}:ovp.medialive:command:stop-channel`);
   const nodeRef = NodeRef.fromNode(node).toString();
   const mediaLiveData = {};
+  useRaven({ node, nodeRef });
```
efjacobson, line 47: @meghedi is this going to send a heartbeat and start collaboration for each of these videos?

src/plugins/ovp/components/livestreams-screen/LivestreamsCard.jsx
meghedi: Done.

src/plugins/ovp/components/livestreams-screen/LivestreamsCard.jsx
rsumilang: @efjacobson yes. this would add visibility to the livestreams page to show anyone editing it, similar to other node pages. let me know if you think it smells.

CleanShot 2024-03-21 at 18 53 22

efjacobson: @rsumilang i like that you can see who is editing it, but the person who is on the livestreams card shouldn't appear to others as if they are editing it (as the start/stop operations have no effect on the associated nodes). seems to me it's essentially like one of the search screen tables where you can see the active editors but aren't one yourself

rsumilang: @efjacobson fair enough

gdbrown: @rsumilang @efjacobson agree, should be like search page... just the collaborators component.

rsumilang:
@meghedi i get taken to the top menu item if i click on the menu expansion arrow while in a smaller window. i think clicking the arrow should only expand the list.

src/components/navbar/index.jsx
```diff
@@ -19,6 +19,7 @@ export default function () {
   const policy = usePolicy();
   const [isActive, setActive] = useState(false);
   const [activeGroup, setActiveGroup] = useState(null);
+  const isSmallScreen = window.innerWidth < 1025;
```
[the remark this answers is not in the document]
meghedi: @efjacobson its all css. 
https://github.com/triniti/cms-js/blob/a2ceece7395af031cbe1a305b32f44e1b472d543/src/assets/styles/components/_navbar.scss#L26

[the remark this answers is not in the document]
efjacobson: @meghedi doesnt it come down to the toggler though? in fact, its weird that the toggle function is being called for both the NavbarToggler and the RouterLink onClicks. they may share some functionality but they are ultimately different eh? and you should be able to use the state of the toggler to determine whether it is a "small screen"

gdbrown:
@meghedi v2.x has been massively updated, please pull in the changes and adjust as needed.  everything must use full paths now, including extensions.

rsumilang:
@meghedi conflicts

gdbrown:
@meghedi merged your livestream stuff into v2.x and made some adjustments.  not sure if you need this branch/or any more.

meghedi:
> @meghedi merged your livestream stuff into v2.x and made some adjustments. not sure if you need this branch/or any more.

@gdbrown, @rsumilang those were the only changes left, css changes were updated by Matt on v2.x before, so I think we don't need this branch anymore.

PR #372, opened 2024-03-30
v2 build updates

rsumilang:
fyi - tests are not included here


gdbrown:
@efjacobson @rsumilang can y'all get this current with v2.x and maybe we get megs changes in too so i can blast it through.

PR #373, opened 2024-04-16
V2.x build vite

rsumilang:
@gdbrown @efjacobson Check it out, vite implementation as a proof of concept. I think everything is working from a high level. I haven't went though each component and clicked on everything but in general all the screens seem to work. I wanted to pause here and check in before going further down the rabbit hole. Let me know if you have any concerns.

---

- Demo application runs on Vite
- Support for environment specific `.env` by `.env.[evironment]`
- Build remains with Babel
- Webpack removed
- CMS components referenced via `@triniti/cms`. Needed for demo to work otherwise we have to build CMS first into `dist` and have the demo reference those file. The caveat with this being we would need to have a separate build watcher on the main src to continuously keep the `dist` updated while active development.


efjacobson:
@rsumilang i know we've discussed it a bit on slack, but can you list here the issues you were having with webpack?

rsumilang:
@efjacobson the issue with webpack involves pathing and transpiling. I have unresolvable errors stating that there wasn't a loader configured for react files in the triniti cms when importing triniti into other projects. One could say thats an issue of the external project and the updates here probably don't solve that and they might be right. I was able to solve a lot of these problems by using vite on my external project and felt that I should spread the love back to the triniti world to reap the benefits. Benefits being a simpler setup and a faster build system.

On a side note, vite pointed out some bad file references that required fixing in this branch. These things could've possibly been the culprit with webpack on importing triniti in my other project. One example being a react file with a .js extension. It might be worth to revisit webpack to see if this branch does indeed fix those errors on external projects but the speed, productivity, and clear errors vite was giving me felt like it would be going backwards so I haven't made time for it yet.

gdbrown:
@rsumilang @efjacobson it's a lot to review, unpack.  i'm not familiar with vite (had investigated bun as a replacement but found it lacking).  my main concern is whether or not the same problems exist in vite, we simply don't see em yet.  i truly hate js packaging and bundling.  the main issue is local dev convenience/simplicity/speed, is this easier?

btw, i killed eslint for a reason, i hate that too, as much as webpack.  my god i hate eslint.

rsumilang:
@gdbrown fwiw, i've gotten way further with our internal app using vite without all the friction that webpack adds. actually, i'm near completion in terms of the build system portion of it. we can keep this discussion open and I can circle back here to let you know hows things go at the finish line. with webpack, i got a caught up chasing my tail. meanwhile, do you remember what limitations bun had? i can check where vite stands with them.

efjacobson:
> btw, i killed eslint for a reason, i hate that too, as much as webpack. my god i hate eslint.

fwiw we're on [semistandard](https://github.com/standard/semistandard) these days which has been pretty much painless

gdbrown:
@rsumilang @efjacobson do y'all know if the babel alias overwriting works at site level with this change?  how we override some screens/components with the babel-plugin-module-resolver.

baking in the `@triniti/cms` in this code beforehand, vs letting node/babel resolve it once it's a package might be funky.  probably won't find out until we publish it and try it.

gdbrown:
@rsumilang @efjacobson should be just wrap up a beta release and see what does and doesn't work at site level?  is it done enough?  seems close.

gdbrown:
@efjacobson @rsumilang i'm in the same boat.  it's the devil we know(ish).  macromedia flash was simpler, we should go back.  maybe java applets.

gdbrown:
@rsumilang can we get rid of the need to have `@triniti/cms` pathing in the app itself.  seems odd since it's within its own package.  also, change the type to module in root package.json and see if it still works.

gdbrown:
@rsumilang @efjacobson it needs to be published as esm/module but these extra steps are really gnarly.  trying to get it to work with just babel and no monorepo (which is imo another mega complexity we don't need here).  trying a few things, we'll see if it works.

PR #402, opened 2024-07-09
Clickable Row

rsumilang:
The PR is an example of the assets search screen as having each row clickable.

- Added NodePolyfillPlugin because MimeTypes required a path.extname polyfill


src/plugins/dam/components/search-assets-screen/index.js
```diff
@@ -98,10 +100,10 @@ function SearchAssetsScreen(props) {
                 const transcodingStatus = node.get('transcoding_status');
 
                 return (
-                  <tr key={`${node.get('_id')}`} className={`status-${node.get('status')}`}>
+                  <tr key={`${node.get('_id')}`} className={`status-${node.get('status')}`} style={{cursor: 'pointer'}}>
```
gdbrown, line 103: @rsumilang use the `cursor-pointer` utility class in className instead.

src/plugins/dam/components/search-assets-screen/index.js
```diff
@@ -98,10 +100,10 @@ function SearchAssetsScreen(props) {
...
                     <td><Input type="checkbox" onChange={() => batch.toggle(node)} checked={batch.has(node)} /></td>
                     <td className="text-center"><AssetIcon id={node.get('_id')} /></td>
-                    <td>
+                    <td onClick={() => navigate(nodeUrl(node, 'view'))}>
```
gdbrown, line 106: @rsumilang can we make an onClick function and put it on the row (one time).  accept the event and inspect the target to see if it has a data-ignore-row-click.  then put the data-ignore-row-click on the td we don't need a handler for every td?

gdbrown (changes requested):

src/plugins/dam/components/search-assets-screen/index.js
rsumilang: @gdbrown there seems to be a bug in chrome when using the class name instead. the `cursor: pointer` setting is inherited down to each `td` but the cursor does not change. only changes when using the class if applied to each individual `td`.

src/plugins/dam/components/search-assets-screen/index.js
rsumilang: ok check it out

src/plugins/dam/components/search-assets-screen/index.js
```diff
@@ -15,6 +16,7 @@ import SearchForm from '@triniti/cms/plugins/dam/components/search-assets-screen
 import BatchOperationsCard from '@triniti/cms/plugins/dam/components/search-assets-screen/BatchOperationsCard.js';
 import useBatch from '@triniti/cms/plugins/ncr/components/useBatch.js';
 import AssetIcon from '@triniti/cms/plugins/dam/components/asset-icon/index.js';
+import nodeClickNavigator from '@triniti/cms/utils/nodeClickNavigator.js';
```
gdbrown, line 19: @rsumilang can you call it `createRowClickHandler.js` and the data attribute to `data-ignore-row-click`

src/plugins/dam/components/search-assets-screen/index.js
```diff
@@ -96,11 +99,12 @@ function SearchAssetsScreen(props) {
                 const canUpdate = policy.isGranted(`${schema.getQName()}:update`);
                 const seq = node.get('gallery_seq');
                 const transcodingStatus = node.get('transcoding_status');
+                const handleNodeClick = nodeClickNavigator(navigate, node);
```
gdbrown, line 102: @rsumilang call it `const handleRowClick = createRowClickHandler(navigate, node);`

src/plugins/dam/components/search-assets-screen/index.js
```diff
@@ -96,11 +99,12 @@ function SearchAssetsScreen(props) {
...
 
                 return (
-                  <tr key={`${node.get('_id')}`} className={`status-${node.get('status')}`}>
-                    <td><Input type="checkbox" onChange={() => batch.toggle(node)} checked={batch.has(node)} /></td>
-                    <td className="text-center"><AssetIcon id={node.get('_id')} /></td>
+                  <tr key={`${node.get('_id')}`} className={`status-${node.get('status')}`} style={{cursor: 'pointer'}} onClick={handleNodeClick}>
```
gdbrown, line 105: @rsumilang you won't need this style, i tested a fix on sloth branch and it seems to work in chrome.
```
<tr key={`${node.get('_id')}`} className={`cursor-pointer status-${node.get('status')}`}>
```

src/plugins/dam/components/search-assets-screen/index.js
```diff
@@ -96,11 +99,12 @@ function SearchAssetsScreen(props) {
...
+                    <td data-ignore-node-click><Input type="checkbox" onChange={() => batch.toggle(node)} checked={batch.has(node)} /></td>
```
gdbrown, line 106: @rsumilang use `data-ignore-row-click`

gdbrown (changes requested):

src/plugins/dam/components/search-assets-screen/index.js
```diff
@@ -96,11 +99,12 @@ function SearchAssetsScreen(props) {
                 const canUpdate = policy.isGranted(`${schema.getQName()}:update`);
                 const seq = node.get('gallery_seq');
                 const transcodingStatus = node.get('transcoding_status');
+                const handleRowClick = createRowClickHandler(navigate, node);
 
                 return (
-                  <tr key={`${node.get('_id')}`} className={`status-${node.get('status')}`}>
-                    <td><Input type="checkbox" onChange={() => batch.toggle(node)} checked={batch.has(node)} /></td>
-                    <td className="text-center"><AssetIcon id={node.get('_id')} /></td>
+                  <tr key={`${node.get('_id')}`} className={`status-${node.get('status')}`} onClick={handleRowClick}>
```
gdbrown, line 105: @rsumilang can you add that css class as well?  maximum spice

PR #411, opened 2024-07-12
ovp missing

rsumilang:


gdbrown (changes requested):
@rsumilang these are there already.  https://github.com/triniti/cms-js/blob/v2.x/src/plugins/ovp/components/video-screen/AssetsTab.js

gdbrown (changes requested):
@rsumilang load it up and see it... it's all there https://github.com/triniti/cms-js/blob/v2.x/src/plugins/ovp/components/video-screen/SyndicationCard.js

gdbrown:
@rsumilang note that we left of kaltura on purpose, we don't use them any more

PR #414, opened 2024-07-13
clickity asset table list?

rsumilang:
do you think this is more consistent to the asset search pages?


gdbrown (commented):
@rsumilang i dunno, feels less intuitive in this scenario (in modals).  i prefer to the select button.

PR #426, opened 2024-07-17
headline fragments

rsumilang:


src/plugins/news/components/headline-fragments-card/index.js
```diff
@@ -0,0 +1,52 @@
...
+const OPTIONS_SIZE = [
+  { label: 'XL', value: 1 },
+  { label: 'L', value: 2 },
+  { label: 'M', value: 3 },
+  { label: 'S', value: 4 },
+  { label: 'XS', value: 5 },
+  { label: 'XXS', value: 6 },
+];
+
+export default function HeadlineFragmentsField(props) {
```
gdbrown, line 22: @rsumilang match function name to component.

src/plugins/news/components/article-screen/StoryTab.js
```diff
@@ -4,6 +4,7 @@ import SlugField from '@triniti/cms/plugins/ncr/components/slug-field/index.js';
 import BlocksmithField from '@triniti/cms/components/blocksmith-field/index.js';
 import { DatePickerField, KeyValuesField, NumberField, SwitchField, TextField } from '@triniti/cms/components/index.js';
 import PicklistField from '@triniti/cms/plugins/sys/components/picklist-field/index.js';
+import HeadlineFragmentsCard from '../headline-fragments-card/index.js';
```
gdbrown, line 7: @rsumilang use full path name like all the other things

src/plugins/news/components/headline-fragments-card/index.js
```diff
@@ -0,0 +1,52 @@
...
+  const { pbjName, groupClassName = '' } = props;
```
gdbrown, line 23: @rsumilang is this working?  seems like we're missing a lot to make this work

gdbrown (changes requested):
@rsumilang seems incomplete.  i haven't booted the site with it but i can tell we're missing some spice

src/plugins/news/components/headline-fragments-card/index.js
rsumilang: CleanShot 2024-07-17 at 08 45 23

 Seems like it

gdbrown: @rsumilang what about with new articles?  also, try not setting size/style or text on some and save.  note that pbjName refers to the schema's name, whereas "name" is for final form.  e.g. https://github.com/triniti/cms-js/blob/v2.x/src/plugins/curator/components/widget-screen/SearchArticlesRequestFields.js#L26

gdbrown: @rsumilang we don't have a nested pbj (like poll answers or widget datasource), we have three separate fields that are arrays that when managed should always be the same length/count.

src/utils/FormMarshaler.js
```diff
@@ -57,7 +57,7 @@ export default class FormMarshaler extends ObjectSerializer {
       if (!payload.hf) {
         payload.hf = (new Array(3)).fill('');
         payload.hf_styles = ['uppercase', 'uppercase', 'uppercase'];
-        payload.hf_size = [3, 1, 1];
+        payload.hf_sizes = [3, 1, 1];
```
gdbrown, line 60: @rsumilang we can't change core machinery of form processing/marshalling/etc. for specific fields.  gnarles barkley.  this needs to be contained within the component that provides the fields.  for example, creating some unmanaged fields and using the field registration with final form api and having a "beforeSubmit" that converts the value before form submission.

this feature must be self contained tho.

PR #429, opened 2024-07-17
Widget Picker Items full column size

rsumilang:


gdbrown:
@rsumilang not sure you need a param here, i assume this is because the sortable on picker is wonky, one could assume this based on sortable being used.  however, i plan to refactor the sortable out of the select itself.  feels wonky af rn.

gdbrown:
@rsumilang yeah, i just tried this fix out and although it's better it still drives me nuts how it works when the picker tries to load data when you click it and then while it's open you have to try and sort them. my plan was to move the sortable strategy out of the select entirely (above the select, similar to last version).  then the picker merely adds to that list and sort/remove happens in the sortable above it.

rsumilang:
@gdbrown you can sort the list without having a loaded data list open. you can go straight to an item and drag it. notice the cursor change to a hand when you hover over it.

gdbrown:
@rsumilang I'm aware, i see the hand and what not.  on firefox though, it always opens the search when you click it

rsumilang:
> @rsumilang I'm aware, i see the hand and what not. on firefox though, it always opens the search when you click it

Can we take this patch in until I refactor is done?

rsumilang:
@gdbrown done

gdbrown:
@rsumilang can you leave the select-with-image intact, it's used elsewhere.  select-with-image also might mean something different than select-stacked.  similar but not same.

gdbrown:
@rsumilang i put it in chonks and merged it

PR #448, opened 2024-08-07
Update uriTemplates.js

rsumilang:


demo/src/config/uriTemplates.js
```diff
@@ -28,6 +28,7 @@ registerGlobals({ web_base_url: webBaseUrl });
 window.DAM_BASE_URL = damBaseUrl;
 window.IMAGE_BASE_URL = imageBaseUrl;
 window.VIDEO_BASE_URL = ovpBaseUrl;
+window.WEB_BASE_URL = webBaseUrl;
```
gdbrown, line 31: @rsumilang what's this for?

gdbrown: @rsumilang closing this, there's no use case for that window global.  the others are there for the damUrl.js which uses a different strategy to generate urls

rsumilang: @gdbrown 
CleanShot 2024-08-07 at 17 31 13

rsumilang: @gdbrown  CleanShot 2024-08-07 at 17 34 12

gdbrown: @rsumilang site specific, your uriTemplates file can certainly do that but that banner is not present here.

rsumilang: @gdbrown would you like my proposal in the form of a PR?

gdbrown: @rsumilang send a more complete screenshot (over slack if oyu want) and some code example of status is doing, is that just a link?

also, is this different per environment?  how so if so why so.

PR #462, opened 2025-03-06
gallery support from asset picker

rsumilang:
Adding support to choose primary image from gallery in the block editor.


gdbrown (changes requested):
@rsumilang this can be done without a new tab or new props, use the searchEnricher prop to customize the request.  this would force the search to always return images that are in the gallery (not sure if this is ideal).  also note, the search form already allows you to filter by gallery.  maybe it just needs to be set the first time so you can change it if you want.  you'd pass a searchEnricher from the gallery block modal, it should pass through.

rsumilang:
@gdbrown i don't want to lock the user in the gallery. let me know if the new changes i made are sufficient or if there is a better way

gdbrown (approved):

PR #464, opened 2025-06-03
Docs

rsumilang:


gdbrown:
@rsumilang have you tried using copilot, cursor or codex? just ask a model to explain or write new code.  it's not bad.

rsumilang:
@gdbrown yeah they're not bad. however i think bringing documenting to the forefront for the new core functions is helpful for our developers as an entry point.

rsumilang:
@gdbrown I want to mention that there is also a fix here for the create article modal. Would you rather me separate that out into another PR or zap the documentation?

rsumilang:
@efjacobson @gdbrown our qa team found 2 more issues that this ticket now addresses.

1. the question field in polls modal and form is now required.
2. there was a search request error on widgets forms when there was no data source selected. this is fixed now.

src/plugins/curator/components/widget-screen/DataSourceTab.js
```diff
@@ -66,7 +66,6 @@ export default function DataSourceTab(props) {
             label="Content Type"
             options={options}
             isClearable={false}
```
gdbrown, line 68: @rsumilang that field is required by schema though, in what scenario could/is this blank?

efjacobson: @gdbrown it is not required - https://github.com/triniti/schemas/blob/563218e8cdd37309d39b606b33a1d197f94bd9c6/schemas/triniti/curator/mixin/widget-has-search-request/latest.xml#L18-L22

we've got it up and running where you can make an alert widget with no search request (because its not required) but then once made, if you try to edit any field it starts yelling at you for not having set a search_request

rsumilang: @gdbrown its blank if you use the create widget modal to create a new widget.

gdbrown: @rsumilang @efjacobson oic, blank on the parent object potentially.  putrid.  i'll roll it into the sausage.

gdbrown (approved):

PR #465, opened 2025-06-24
v1.2.1

rsumilang:
Fixed react-final-form v7 compatibility where FieldArray field values could be undefined during initial render, causing destructuring errors in SortableAnswer and SortableSlot components.

### DIFF
--- a/CHANGELOG-1.x.md
+++ b/CHANGELOG-1.x.md
@@ -1,6 +1,10 @@
 # CHANGELOG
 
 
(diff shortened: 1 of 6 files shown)


gdbrown:
@rsumilang is it possible to just return early when it's undefined rather than make these mods to cover all those scenarios?

rsumilang:
@gdbrown done

gdbrown (approved):

PR #467, opened 2025-07-26
v1.2.3

joshlevy:
* Fixed batch-operations-card to use the value from enums for conditional comparisons rather than the enum itself.


joshlevy:
Closing this pr because issue was already addressed.

PR #470, opened 2025-08-07
Swipe disappearing after save

hunyoboy:
* Fixed single select field not hydrating data.

### DIFF
--- a/CHANGELOG-1.x.md
+++ b/CHANGELOG-1.x.md
@@ -3,6 +3,7 @@
 
 ## vN.N.N
 * Use `Swal.DismissReason` enum in `receiveEnvelope.js` and `Authorizer.js`
+* Fix Swipes disappearing after save
 
 
 ## v1.2.2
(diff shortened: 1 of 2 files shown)


CHANGELOG-1.x.md
```diff
@@ -3,6 +3,7 @@
 
 ## vN.N.N
 * Use `Swal.DismissReason` enum in `receiveEnvelope.js` and `Authorizer.js`
+* Fix Swipes disappearing after save
```
efjacobson: @hunyoboy can you make this a bit more descriptive/accurate? what we're really talking about here is the form fields are not hydrating properly, right?

hunyoboy: @efjacobson updated

CHANGELOG-1.x.md
```diff
@@ -3,6 +3,7 @@
 
 ## vN.N.N
 * Use `Swal.DismissReason` enum in `receiveEnvelope.js` and `Authorizer.js`
+* picklist-field :: Fixed single select field not hydrating data after a successful save.
```
efjacobson: @hunyoboy is this specific to single select fields that are using picklists? seems like it could happen to any single select field, given the right configuration

hunyoboy: @efjacobson that's true. i'm thinking along the lines of `location of issue found : description`.

efjacobson (approved):

PR #471, opened 2025-08-07
Publish Immediately

guizzettic:
* Publish Immediately

### DIFF
--- a/CHANGELOG-1.x.md
+++ b/CHANGELOG-1.x.md
@@ -3,6 +3,7 @@
 
 ## vN.N.N
 * Use `Swal.DismissReason` enum in `receiveEnvelope.js` and `Authorizer.js`
+* Fixed publish immediately to use server time instead of client time
 
 
 ## v1.2.2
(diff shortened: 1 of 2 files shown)


guizzettic:
This PR fixes a regression where articles are scheduled instead of publishing them immediately when a user's system clock is incorrect. We discovered this issue when a user reported articles were showing "Scheduled" status although he was publishing them, and we noticed his device's clock had drifted 53 seconds ahead of actual time. Hypothetically, depending on who this occurred to, this could be a major issue that would incur untold wrath.

The root cause is that `PublishForm` currently sends out the client time when "Publish" is selected, causing the server to treat this as a scheduled publish if the client's clock differs from server time. This breaks the expectation that "publish immediately" should be instantaneous.

The fix ensures that when users select "Publish", we explicitly pass null for the `publish_at` param, deferring to the server's timestamp.

efjacobson:
@gdbrown lmk what you think of this. we've got a batch of changes coming that i'm gonna roll into one big pr for you but this one seemed like a bit of an outlier so @guizzettic got you involved early

rsumilang:
@guizzettic conflicts

efjacobson (approved):

PR #472, opened 2025-08-08
Filename preview galleries

meghedi:
Add filename preview on image hover in galleries


src/assets/styles/components/_backdrop-image.scss
efjacobson, line 1: @meghedi this doesnt seem like the right place for these new styles. your new react code is not a backdrop-image, right?

src/plugins/curator/components/gallery-screen/images-tab/SortableImage.js
```diff
@@ -64,6 +64,11 @@ export default function SortableImage(props) {
         {(isHovering || isSelected) && (
           <div className="position-absolute w-100 h-100 bg-opacity-50 bg-black"></div>
         )}
+        <div className="gallery-image-title-overlay">
+          <div className="text-white text-center small text-truncate">
+            {image.get('title')}
+          </div>
+        </div>
```
efjacobson: @meghedi seems like you should use `isHovering` here since it's available and in use already

efjacobson (approved):

PR #474, opened 2025-08-20
Widget DataSource

rsumilang:
* curator :: Widget datasource fixes. Clearable datasource, vendor schema fix, and adding `page` back.

### DIFF
--- a/src/plugins/curator/components/widget-screen/DataSourceTab.js
+++ b/src/plugins/curator/components/widget-screen/DataSourceTab.js
@@ -74,7 +74,6 @@ export default function DataSourceTab(props) {
             name="search_request._schema"
             label="Content Type"
             options={options}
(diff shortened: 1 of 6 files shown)


src/plugins/curator/components/widget-screen/DataSourceTab.js
```diff
@@ -65,7 +74,7 @@ export default function DataSourceTab(props) {
             name="search_request._schema"
             label="Content Type"
             options={options}
-            isClearable={false}
+            isClearable={true}
```
efjacobson: @rsumilang nit: should be able to to just put `isClearable` with no value and it will be true

efjacobson:
just needs a ticket for tracking (and conflict resolution) and this will be good to go

efjacobson (approved):

PR #478, opened 2025-09-06
iframe embed support

rsumilang:
* blocksmith :: adding support for embed codes in iframe src input.


efjacobson:
@rsumilang lmk when the unit test we discussed is added

rsumilang:
@efjacobson done with unit test

efjacobson (approved):

PR #480, opened 2025-09-10
156 filter selector

meghedi:
tmz-apps/toofab-cms#156


src/plugins/curator/components/gallery-picker-field/index.js
```diff
@@ -1,11 +1,28 @@
...
+    
+    if (statuses) {
+      initialData.statuses = statuses;
+    }
+    
+    return withRequest(NodePickerField, 'triniti:curator:request:search-galleries-request', {
+      channel: 'picker',
+      initialData
+    });
+  }, [JSON.stringify(statuses)]);
```
efjacobson: @meghedi are you sure you need to stringify here? the statuses that are passed in should have consistent references

src/plugins/curator/components/gallery-picker-field/index.js
```diff
@@ -1,11 +1,28 @@
...
 
-export default withRequest(NodePickerField, 'triniti:curator:request:search-galleries-request', {
-  channel: 'picker',
-  initialData: {
-    sort: SearchGalleriesSort.ORDER_DATE_DESC.getValue(),
-    autocomplete: true,
-  }
-});
+const GalleryPickerFieldWithStatus = (props) => {
+  const { statuses, ...otherProps } = props;
```
efjacobson: @meghedi nit: usually we call the other props `...rest`

efjacobson (commented):
@meghedi did you consider doing this work at the NodePickerField level?

src/plugins/curator/components/gallery-picker-field/index.js
meghedi: @efjacobson if I remove the brackets and remove the JSON.stringify, when no statuses are passed as props, it gets stuck on loading. I updated the code to use join instead.

src/blocksmith/components/article-block-modal/ArticlePickerField.js
```diff
@@ -0,0 +1,16 @@
...
+const ArticleBlockArticlePickerField = (props) => {
+    return (
+        <ArticlePickerField
+            {...props}
+            statuses={props.statuses}
+        />
+    );
+};
+
+export default ArticleBlockArticlePickerField;
```
efjacobson: @meghedi keep it simple. this is all you need:


```javascript
import ArticlePickerField from '@triniti/cms/plugins/news/components/article-picker-field/index.js';

export default ArticlePickerField;
```

src/plugins/curator/components/gallery-picker-field/index.js
```diff
@@ -1,11 +1,28 @@
...
+    return withRequest(NodePickerField, 'triniti:curator:request:search-galleries-request', {
+      channel: 'picker',
+      initialData
+    });
+  }, statuses || []);
+ 
+  return <GalleryPickerField {...rest} />;
+};
+
+export default GalleryPickerFieldWithStatus;
```
efjacobson: @meghedi the more i think about this the more i think it is mostly unnecessary
- `useMemo` is generally not used for components. rather `memo` is used for components that have props that are the result of `useMemo`
- that said, have you actually noticed any issues from _not_ using `useMemo`? i can't imagine the performance hit would be much, so if nothing is funky with the request then i don't think we need memoization here at all

this works and doesn't make assumptions about what the vendor may want to use for `initialData`:

```javascript
export default ({ initialData, ...rest }) => withRequest(
  NodePickerField,
  'triniti:curator:request:search-galleries-request',
  {
    channel: 'picker',
    initialData: {
      sort: SearchGalleriesSort.ORDER_DATE_DESC.getValue(),
      autocomplete: true,
      ...initialData
    }
})(rest);

// elsewhere...

<GalleryPickerField name="node_ref" label="Gallery" required initialData={{ statuses: [NodeStatus.PUBLISHED] }}/>
```

it also leaves it up to the vendor to decide how they want to handle prop equality for `initialData`, which is a good thing. the only weirdness i see at the moment is that it feels like the `initialData` prop here should actually be called `resolverConfig` or something because that is where is is used. but that opens up a whole other can of dynamic prop decoration worms

---

actually, this is what it would be with `resolverConfig`, and it's not the grossest thing i've ever seen. thoughts?

```javascript
export default ({ resolverConfig = {}, ...rest }) => {
  const config = {
    channel: resolverConfig.channel || 'picker',
    initialData: {
      sort: SearchGalleriesSort.ORDER_DATE_DESC.getValue(),
      autocomplete: true,
      ...(resolverConfig.initialData || {})
    }
  };
  if (resolverConfig.persist !== undefined) {
    config.persist = resolverConfig.persist;
  }
  return withRequest(NodePickerField,'triniti:curator:request:search-galleries-request', config)(rest);
};

// elsewhere...

<GalleryPickerField name="node_ref" label="Gallery" required resolverConfig={{ initialData: { statuses: [NodeStatus.PUBLISHED] } }}/>
```

meghedi: I think this approach is much better and clean. I made the changes, it works. Thank you.

src/blocksmith/components/article-block-modal/ArticlePickerField.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * Article Picker Field for Article Block Modal
```
efjacobson: @meghedi this is not all that useful of a comment - why don't you mention that these files exists to be overriden?

efjacobson (approved):

PR #481, opened 2025-09-10
Add save close btn

tbone1020:


src/components/save-button-dropdown/index.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import NodeStatus from '@gdbots/schemas/gdbots/ncr/enums/NodeStatus.js';
+import { DropdownMenu, DropdownToggle, UncontrolledDropdown } from 'reactstrap';
+import { ActionButton, Icon } from '@triniti/cms/components/index.js';
+
+export default function SaveButtonDropDown (props) {
+  const { delegate, formState, node, isRefreshing, qname, policy } = props;
```
hunyoboy: @tbone1020 you really don't need the entire delegate. why not just an onClick prop?

src/components/save-button-dropdown/index.js
```diff
@@ -0,0 +1,54 @@
...
+
+  const { dirty, hasSubmitErrors, submitting, valid } = formState;
+  const submitDisabled = submitting || isRefreshing || !dirty || (!valid && !hasSubmitErrors);
+  const canPublish = policy.isGranted(`${qname}:publish`);
+  const isPublished = node.get('status') === NodeStatus.PUBLISHED;
```
hunyoboy: @tbone1020 nit: can be `NodeStatus.PUBLISHED.getValue()`

src/components/save-button-dropdown/index.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import NodeStatus from '@gdbots/schemas/gdbots/ncr/enums/NodeStatus.js';
+import { DropdownMenu, DropdownToggle, UncontrolledDropdown } from 'reactstrap';
+import { ActionButton, Icon } from '@triniti/cms/components/index.js';
+
+export default function SaveButtonDropDown (props) {
```
hunyoboy: @tbone1020 this component's feature is very ncr-specific ( save/publish nodes ). this should be an ncr component.

tbone1020: Moved to ncr/ folder

src/components/save-button-dropdown/index.js
tbone1020: Done

src/components/save-button-dropdown/index.js
tbone1020: @hunyoboy yeah we definitely don't need all of delegate. I only passed in `handleSave`. Is that what you meant?

src/components/save-button-dropdown/index.js
hunyoboy: @tbone1020 i'm thinking that we can override this component on the vendor level. the functionality here should stay the same so it will not affect other vendors. something like:

```javascript
export default function SaveNodeButton (props) {
  const { disabled, nodeRef, onClick: handleClick  } = props;
  const policy = usePolicy();
  const canSave = policy.isGranted(`${NodeRef.fromString(nodeRef).qname}:update`);

  return (
    canSave ?
      <ActionButton
        text="Save"
        onClick={handleClick}
        disabled={disabled}
        icon="save-diskette"
        color="primary"
     /> : null
  );
};
```

let me know what you think.

tbone1020: @hunyoboy so revert my changes of moving the code to the ncr/ folder changes then override it at the vendor level?

hunyoboy: @tbone1020 ncr folder is fine since this is a ncr component. i'm thinking your code with save/publish or save/close functionality  shouldn't be in here. it should be on the vendor level to override this file. let's also call this `SaveNodeButton`.  ref. my sample code above and how they do it here - https://github.com/triniti/cms-js/commit/c62759110ac1399e3af15c09abd4a325dda10181

tbone1020:
@hunyoboy reverted the buttons functionality back. After some thought, I think it would be better to pass in `node` instead of `nodeRef`. This way we're not using `const { node } = useNode(nodeRef);` just to grab something that can be passed in.

rsumilang:
> @hunyoboy reverted the buttons functionality back. After some thought, I think it would be better to pass in `node` instead of `nodeRef`. This way we're not using `const { node } = useNode(nodeRef);` just to grab something that can be passed in.

@tbone1020 the pattern in this system is to pass the nodeRef. this is lighter weight and components can choose when to load nodes.

src/plugins/iam/components/user-screen/index.js
```diff
@@ -62,12 +63,10 @@ function UserScreen(props) {
...
           {canUpdate && (
             <>
-              <ActionButton
-                text="Save"
+              <SaveNodeButton 
                 onClick={delegate.handleSave}
                 disabled={submitDisabled}
-                icon="save-diskette"
-                color="primary"
+                nodeRef={nodeRef}
```
hunyoboy: @tbone1020 isn't it odd that we passed a `nodeRef` here but we're not actually using it inside the component? it's ok for those who know why we do this ( for overriding ). but for a reader that doesn't know the purpose, this can look like it's passed by mistake.

src/plugins/ncr/components/save-node-button/index.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { ActionButton } from '@triniti/cms/components/index.js';
+
+export default function SaveNodeButton (props) {
+  const { disabled, onClick: handleSave } = props;
```
hunyoboy: @tbone1020 why not use the `nodeRef` to determine if user is allowed to save or not? at least, the users/callers of this component will not need to worry about that logic. it also justifies why we extracted it to be a ncr component other than to make it overridable.

src/plugins/ncr/components/with-node-screen/useDelegate.js
```diff
@@ -201,17 +214,23 @@ export default (props) => {
...
   delegate.handleSave = async (event) => {
     const action = event.target?.value || 'save';
+    if (action === 'save-and-publish' && !(await okayToPublish())) {
+      return;
+    }
+
     delegate.handleSubmit = async (values) => {
       try {
-        const ref = NodeRef.fromString(nodeRef);
+        const ref = NodeRef.fromString(nodeRef)
```
hunyoboy: @tbone1020 missing semicolon

src/plugins/iam/components/user-screen/index.js
tbone1020: @hunyoboy yeah that makes sense but if we're only going to overwrite the button component then we need to pass in the nodeRef?

src/plugins/ncr/components/save-node-button/index.js
tbone1020: @hunyoboy We can definitely do that way but there is already a `canUpdate` in the above component. I know we talked about moving the can-save check into the button component but wouldn't it be better to keep it `canUpdate` where it is?

```Javascript
{canUpdate && (
  <>
    <SaveNodeButton 
      onClick={delegate.handleSave}
      disabled={submitDisabled}
      nodeRef={nodeRef}
    />
    <ActionButton
      text={editMode ? 'Enter View Mode' : 'Enter Edit Mode'}
      onClick={delegate.handleSwitchMode}
      disabled={submitting || isRefreshing}
      icon={editMode ? 'eye' : 'edit'}
      color="light"
      outline
    />
  </>
)}
```

src/plugins/ncr/components/with-node-screen/useDelegate.js
tbone1020: Added missing semicolon

src/plugins/iam/components/user-screen/index.js
hunyoboy: @tbone1020 yes we only pass the nodeRef. that's all you need to achieve the save-publish/save-close functionality on vendor level.

however, my concern here is this. someone new works on cms-js repo. that new person may think that we passed the `nodeRef` by mistake here since it's not used inside the component. so he/she may end up deleting the `nodeRef` as a refactor. that will break your update on the vendor level if you pull it. so to prevent that from happening, it's either you put a comment explaining in great detail why we are passing this unused `nodeRef` or just use it inside the component.

src/plugins/ncr/components/save-node-button/index.js
hunyoboy: @tbone1020 don't see anything wrong like this

```javascript
 <SaveNodeButton 
      onClick={delegate.handleSave}
      disabled={submitDisabled}
      nodeRef={nodeRef}
 />
{canUpdate && 
<ActionButton...
```
it just shows `saveNodeButton` has the answer for the "can I update?" question.

tbone1020: @hunyoboy done

src/plugins/iam/components/user-screen/index.js
tbone1020: @hunyoboy done

CHANGELOG-1.x.md
```diff
@@ -2,6 +2,7 @@
 
 
 ## vN.N.N
+* Add save & close and save & publish dropdown buttons to all nodes.
```
hunyoboy: @tbone1020 i think your change log needs to be updated to reflect your actual changes.

hunyoboy (commented):
@tbone1020 looks good. just need to update your current change log.

CHANGELOG-1.x.md
tbone1020: @hunyoboy done.

PR #490, opened 2025-10-03
1482 insert block refinement

mtakla16:


efjacobson:
~looking pretty good to me, what do you think @rsumilang?~

actually, its concerning that there are two places now where InsertBlockButtons may be rendered. is there a way to consolidate that? @mtakla16 

what really is the difference between the insert block plugin and the hover insert buttons? are they not the same thing?

mtakla16:
@efjacobson the InsertBlockButtons component is used in three places:
1. HoverInsertButtons - this is specifically for adding a new block at the top of the article
2. InsertBlockPlugin - used for adding a new block directly below any text-block
3. withBlockPreview - used to add a new block directly underneath any other type of block

It'd probably be a low lift to make InsertBlockPlugin work for all block types so that we don't need InsertBlockButtons in withBlockPreview, but replacing HoverInsertButtons with InsertBlockPlugin would definitely be a little more complicated.

efjacobson:
> It'd probably be a low lift to make InsertBlockPlugin work for all block types so that we don't need InsertBlockButtons in withBlockPreview, but replacing HoverInsertButtons with InsertBlockPlugin would definitely be a little more complicated.

@mtakla16 makes sense, if we leave HoverInsertButtons how they are, how much time do you think you would need to consolidate the rest?

src/blocksmith/components/insert-block-buttons/index.js
```diff
@@ -2,9 +2,9 @@ import React from 'react';
 import { Button } from 'reactstrap';
 import { Icon } from '@triniti/cms/components/index.js';
 
-export default function InsertBlockButtons({ onInsertTextBlock, onInsertBlock, className = 'insert-block-buttons' }) {
+export default function InsertBlockButtons({ onInsertTextBlock, onInsertBlock, className = 'insert-block-buttons', style = {} }) {
   return (
-    <div className={className}>
+    <div className={className} onClick={onInsertTextBlock} style={style}>
```
rsumilang: Is this onClick necessary?

src/blocksmith/components/insert-block-buttons/index.js
```diff
@@ -2,9 +2,9 @@ import React from 'react';
 import { Button } from 'reactstrap';
 import { Icon } from '@triniti/cms/components/index.js';
 
-export default function InsertBlockButtons({ onInsertTextBlock, onInsertBlock, className = 'insert-block-buttons' }) {
+export default function InsertBlockButtons({ onInsertTextBlock, onInsertBlock, className = 'insert-block-buttons', style = {} }) {
```
rsumilang: FYI - We usually alias onFunctions to handleFunctions.

src/blocksmith/plugins/InsertBlockPlugin.js
```diff
@@ -0,0 +1,63 @@
...
+export default function InsertBlockPlugin() {
+  const [editor] = useLexicalComposerContext();
+  const [buttonsStyle, setButtonsStyle] = useState({ display: 'none' });
+  const scrollContainer = document.querySelector('.screen-body');
+  const [nodeKey, setNodeKey] = useState()
+
+  const handleMouseEnterNode = (event, _editor, key) => {
+    const paragraphRect = event.target.getBoundingClientRect();
+    const scrollRect = scrollContainer.getBoundingClientRect();
+    // offset by 37, which is the height of the buttons plus the size of the padding between blocks
```
rsumilang: Can we set the number 37 to named constant? 

`const BUTTON_OFFSET = 37; // Height of buttons plus padding`

rsumilang (changes requested):

src/blocksmith/components/insert-block-buttons/index.js
mtakla16: yeah the user should be able add a new block by clicking anywhere between blocks, not just on the buttons

rsumilang:
@mtakla16 

CleanShot 2025-10-08 at 17 28 33

rsumilang:
@mtakla16 
CleanShot 2025-10-08 at 18 57 44

rsumilang:
I think it looks a little weird that the spacing between blocks is different from the spacing between text blocks. Can we please fix that?

Also, it seems more spacing got added around everything else if you compare to my previous screenshots.

CleanShot 2025-10-08 at 20 01 26

rsumilang:
@efjacobson what do you think of the spacing between blocks being a little tighter than text blocks?

do you think they are fine? should be more spaced out to match the text block spacing, or the text block spacings should be tighter like the blocks?

CleanShot 2025-10-09 at 13 32 55

efjacobson:
@rsumilang looking tight enough now?

rsumilang:
@efjacobson looks ok

efjacobson (approved):

PR #491, opened 2025-10-03
Tiered category styling

tbone1020:


src/plugins/ncr/components/node-picker-field/Option.js
```diff
@@ -23,6 +23,7 @@ export default function Option(props) {
   const status = `${node.get('status')}`;
   const schema = node.schema();
   const isPublishable = schema.hasMixin('gdbots:ncr:mixin:publishable');
+  const label = node.has('tags') ? node.get('tags')?.picker_label : null;
```
efjacobson: @tbone1020 fyi you probably only need the ternary or the optional chainer, not both

```javascript
const label = node.get('tags')?.picker_label;

// or

const label = node.has('tags') ? node.get('tags').picker_label : null;
```

efjacobson (approved):

PR #492, opened 2025-10-06
1472 add floating toolbar plugin

guizzettic:


src/blocksmith/plugins/FloatingTextFormatToolbarPlugin.js
```diff
@@ -0,0 +1,403 @@
...
+  isStrikethrough,
+  isHighlight,
+  blockType,
+}) {
+  const popupRef = useRef(null);
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const isBulletList = blockType === 'bullet';
+  const isNumberList = blockType === 'number';
+
+  const updateTextFormatFloatingToolbar = useCallback(() => {
```
efjacobson: @guizzettic the `$` naming convention they use is [intentional](https://lexical.dev/docs/intro#reading-and-updating-editor-state)

src/blocksmith/plugins/FloatingTextFormatToolbarPlugin.js
efjacobson, line 1: @guizzettic can you add a comment somewhere in here that references the one in the playground that we are borrowing from?

src/blocksmith/plugins/FloatingTextFormatToolbarPlugin.js
guizzettic: @efjacobson Thanks for highlighting that! I updated `updatePopup` to use the `$` naming convention 👍

efjacobson: @guizzettic wrong function

> The closure passed to the update or read call is important, and must be synchronous. It's the only place where you have full "lexical" context of the active editor state, and providing you with access to the Editor State's node tree. _We promote using the convention of using $ prefixed functions (such as $getRoot()) to convey that these functions must be called in this context. Attempting to use them outside of a read or update will trigger a runtime error._ (emphasis added) 

ie its not about functions that may _call_ `editor.getEditorState().read(...)`, its about functions that must be called _within_ `editor.getEditorState().read(...)` (or `editor.read(...)` or `editor.update(...)`)

[this](https://github.com/facebook/lexical/blob/0c9e1eb2c8f1fff0852fab3b6a351a0b44bc44cc/packages/lexical-playground/src/plugins/FloatingTextFormatToolbarPlugin/index.tsx#L121) is the function i was talking about (hence why i put my comment where i did). to fully follow the convention i think we'd want to import `getSelectedNode` as `$getSelectedNode`, because it uses a `$` function inside. might even want to rename that file but lets leave that for now

src/blocksmith/plugins/FloatingTextFormatToolbarPlugin.js
```diff
@@ -0,0 +1,407 @@
...
+    isUnderline,
+    isStrikethrough,
+    isHighlight,
+    blockType,
+  };
+}
+
+/**
+ * Based on Lexical's FloatingTextFormatToolbarPlugin
+ * @see https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/plugins/FloatingTextFormatToolbarPlugin/index.tsx
```
efjacobson: @guizzettic link to the file as it was when you borrowed from it please. you can just follow the link in the comment and [press y](https://docs.github.com/en/repositories/working-with-files/using-files/getting-permanent-links-to-files#press-y-to-permalink-to-a-file-in-a-specific-commit)

src/blocksmith/plugins/FloatingTextFormatToolbarPlugin.js
guizzettic: @efjacobson Got it, thanks for the extra clarification I misunderstood. I thought you were sharing it at a highlevel, as I thought `updateTextFormatFloatingToolbar` didn't necessarily qualify but I see now

efjacobson (approved):

PR #498, opened 2025-10-16
Allow slug spaces

meghedi:


efjacobson (changes requested):
@meghedi this is not working when creating an article (you are able to enter spaces, but they do not get converted to dashes)

src/plugins/ncr/components/slug-field/RenameForm.js
```diff
@@ -86,6 +69,11 @@ function RenameForm(props) {
 
   // todo: add inline alert about 404 when renaming a published node
 
+  const formatSlug = (value) => {
+    if (!value || value[value.length-1] === ' ') return value;
+    return trimStart(value).replace(/\s+/g, '-').toLowerCase();
+  };
+
```
efjacobson: @meghedi rather than throwing out `createSlug` entirely, lets try something like this pseudocode

```javascript
if !value
  return value
if value.endsWith(' ')
  // sluggify everything except last character, then add the space back
  return `${createSlug(value.slice(0, -1), withDatedSlug)} `.toLowerCase()
else
  return createSlug(value, withDatedSlug).toLowerCase()

```

i was also gonna say be careful not to throw out all the parsing logic that you deleted, and i see that it is doing the exact same thing for `/` and `-` characters...you should take the same approach

src/plugins/ncr/components/slug-field/RenameForm.js
```diff
@@ -17,7 +17,8 @@ import getFriendlyErrorMessage from '@triniti/cms/plugins/pbjx/utils/getFriendly
 // more restrictive DATED_SLUG_PATTERN than what gdbots/pbj does
 const DATED_SLUG_PATTERN = /^\d{4}\/\d{2}\/\d{2}\/[a-z0-9-]+$/;
 
-const slugValidator = value => isValidSlug(value) ? undefined : 'Only use letters, numbers and dashes.';
+const slugValidator = value => isValidSlug(value) ? undefined : 'Only use letters and numbers.';
```
efjacobson: @meghedi do you need to change this?

src/plugins/ncr/components/slug-field/RenameForm.js
meghedi: updated to use your code similar to parse logic.

efjacobson: @meghedi can you just add the space character to the others in the parse functions that are already there? or does that not work?

efjacobson (commented):
@meghedi looks like the create modal will allow you to submit with a trailing space, and simply remove it before creating the article. can we do a similar thing for the rename form?

meghedi:
> @meghedi looks like the create modal will allow you to submit with a trailing space, and simply remove it before creating the article. can we do a similar thing for the rename form?

@efjacobson added allow trailing space to the rename form.

efjacobson:
@meghedi the create form is now allowing slashes:

Screenshot 2025-10-22 at 11 09 00
Screenshot 2025-10-22 at 11 09 08

src/plugins/news/components/create-article-modal/index.js
```diff
@@ -17,10 +17,29 @@ import trimStart from 'lodash-es/trimStart.js';
...
+const slugValidator = (value) => {
+  if (!value) return undefined;
+  if (value.includes('/')) {
+    if (!isValidDatedSlug(value)) {
+      return 'Expected format YYYY/MM/DD/some-title-here.';
+    }
+  }
+  const trimmedValue = value?.trim().replace(/\s+/g, '-') ?? value;
+  return isValidSlug(trimmedValue, true) ? undefined : 'Only use letters, numbers and dashes.';
+};
```
efjacobson: @meghedi why is this so different from the other one? shouldnt they be essentially the same?

meghedi: @efjacobson the difference is, with the renameForm datedSlugValidator, we should always have a dated slug but here if we don't have it dated, on submit the dated part will be added. 
let me know please, which one is preferred?

efjacobson: @meghedi just so i understand, the flow would be:
1. user opens create article modal
2. user enters title
3. user blurs title and a dated slug is generated
4. user can delete the dated part and still submit the form

is that what you're talking about? as opposed to the rename form which will render an error and not allow submitting the form if you delete the date?

meghedi: yes, that's correct.

efjacobson: @meghedi i dont think that's a workflow we need to support. the date is already added automatically when creating a new article, if they delete it without entering a new one then they can read the error. as you've seen, this is more complicated than it seems. lets not make it any more complicated than it has to be

src/plugins/ncr/components/slug-field/RenameForm.js
```diff
@@ -17,9 +17,14 @@ import getFriendlyErrorMessage from '@triniti/cms/plugins/pbjx/utils/getFriendly
...
+  const trimmedValue = value ? value.trim().replace(/\s+/g, '-') : value;
+  return isValidSlug(trimmedValue) ? undefined : 'Only use letters, numbers and dashes.';
+};
+
 const datedSlugValidator = (value) => {
-  if (isValidSlug(value, true) && DATED_SLUG_PATTERN.test(trimStart(value))) {
+  const trimmedValue = value ? value.trim().replace(/\s+/g, '-') : value;
+  if (isValidSlug(trimmedValue, true) && DATED_SLUG_PATTERN.test(trimStart(trimmedValue))) {
     return undefined;
   }
```
efjacobson, line 27: @meghedi why do you need to do the trim and replace here? seems like you could accomplish the same with [format](https://final-form.org/docs/react-final-form/types/FieldProps#format)

```jsx
format={(value) => value?.trim()}
formatOnBlur
```

src/plugins/news/components/create-article-modal/index.js
```diff
@@ -17,10 +17,27 @@ import trimStart from 'lodash-es/trimStart.js';
...
+const parseDatedSlug = (value) => {
+  let ending = '';
+  if (value && (value.endsWith(' ') || value.endsWith('/') || value.endsWith('-'))) {
+    ending = value[value.length - 1];
+  }
+  const slug = createSlug(value.toLowerCase(), true);
+  return slug ? slug + ending : value;
+};
+
+const slugValidator = (value) => {
```
efjacobson: @meghedi nit: these can either both have non-dated names (because we're in the create article modal which will always be dated) or they can both have dated names, but not one each

src/plugins/ncr/components/slug-field/RenameForm.js
meghedi: I updated the validator, we do not need this `value.trim().replace(/\s+/g, '-')` in the slugValidator, since its being done in parse,
I kept trim() to avoid a validator error from a trailing space if preserved by parse, it’s removed on submit.

meghedi: format did not work, it did not allow to type space.

efjacobson: @meghedi it does if you have `formatOnBlur`

meghedi: @efjacobson you are right, it worked, thanks

efjacobson (approved):

PR #500, opened 2025-10-17
transcoding error visibility

guizzettic:


rsumilang:
@guizzettic please make an adjustment so that the transcoding error on the video page is displayed on all tabs, not just the details.

guizzettic:
@rsumilang Addressed updated UI request to have the Transcoding Error Alert display across the tabs of Videos and Assets pages

rsumilang:
@guizzettic it looks like we lost the spacing:

CleanShot 2025-11-05 at 17 58 06

i think it looks better without caps lock:

CleanShot 2025-11-05 at 17 59 18

guizzettic:
@rsumilang Updated to add the spacing and not use caps lock 
image

src/plugins/dam/components/asset-screen/TranscodingErrorAlert.js
```diff
@@ -11,8 +12,8 @@ function TranscodingErrorAlert(props) {
   }
 
   return (
-    <Alert color="danger" className="mb-3">
-      <strong>TRANSCODING STATUS: {status.getValue().toUpperCase()}</strong>
+    <Alert color="danger" className="my-3">
+      <strong>Transcoding Status: {capitalize(status.getValue())}</strong>
```
rsumilang: you could remove the `.getValue()` here.

src/plugins/ovp/components/video-screen/TranscodingErrorAlert.js
```diff
@@ -34,8 +35,8 @@ function TranscodingErrorAlert(props) {
   const status = asset.get('transcoding_status');
 
   return (
-    <Alert color="danger" className="mb-3">
-      <strong>TRANSCODING STATUS: {status.getValue().toUpperCase()}</strong>
+    <Alert color="danger" className="my-3">
+      <strong>Transcoding Status: {capitalize(status.getValue())}</strong>
```
rsumilang: you could remove the `.getValue()` here.

rsumilang (changes requested):
One more little minor nit. Thanks!

src/plugins/dam/components/asset-screen/ProcessingErrorAlert.js
```diff
@@ -0,0 +1,35 @@
...
+                             transcriptionStatus === TranscriptionStatus.CANCELED;
+
+  if (!transcodingFailed && !transcriptionFailed) {
+    return null;
+  }
+
+  return (
+    <>
+      {transcodingFailed && (
+        <Alert color="danger" className="my-3">
```
rsumilang, line 24: You can make this a couple lines shorter if you replace the empty tags with the Alert tags.

src/plugins/ovp/components/video-screen/ProcessingErrorAlert.js
```diff
@@ -0,0 +1,64 @@
...
+  const transcodingStatus = asset.get('transcoding_status', TranscodingStatus.UNKNOWN);
+  const transcriptionStatus = asset.get('transcription_status', TranscriptionStatus.UNKNOWN);
+
+  const transcodingFailed = transcodingStatus === TranscodingStatus.FAILED || 
+                           transcodingStatus === TranscodingStatus.CANCELED;
+  const transcriptionFailed = transcriptionStatus === TranscriptionStatus.FAILED || 
+                             transcriptionStatus === TranscriptionStatus.CANCELED;
+
+  return (
+    <>
```
rsumilang, line 45: Same thing here.

rsumilang (changes requested):

src/plugins/dam/components/asset-screen/ProcessingErrorAlert.js
guizzettic: @rsumilang Ok so would you prefer one alert and have both error codes there? The reason for this is i was having an alert show up for each error code.

rsumilang: @guizzettic ah i see. that make sense. lets keep it as is.

rsumilang (approved):

PR #501, opened 2025-10-21
grid view for gallery and linked assets

guizzettic:


mtakla16 (approved):

src/plugins/dam/utils/getAssetDisplayView.js
```diff
@@ -0,0 +1,20 @@
+import AssetTable from '@triniti/cms/plugins/dam/components/linked-assets-card/AssetTable.js';
+import ImageGrid from '@triniti/cms/plugins/dam/components/asset-picker-field/ImageGrid.js';
+import AssetCardGrid from '@triniti/cms/plugins/dam/components/asset-picker-field/AssetCardGrid.js';
+
+export default function getAssetDisplayView(displayView, nodes = [], type) {
```
rsumilang: I think this could be a component and placed in @triniti/cms/plugins/dam/components/asset-picker-field/ with the rest of the components it renders. You could then call it `AssetDisplayer` or `AssetViewer` or something.

src/plugins/dam/utils/getAssetDisplayView.js
```diff
@@ -0,0 +1,20 @@
...
+  if (displayView === 'asset-grid') {
+    return AssetCardGrid;
+  }
+
+  if (displayView === 'table') {
```
rsumilang: maybe this should match the component name and use `asset-table`?

src/plugins/dam/utils/getAssetDisplayView.js
```diff
@@ -0,0 +1,20 @@
+import AssetTable from '@triniti/cms/plugins/dam/components/linked-assets-card/AssetTable.js';
+import ImageGrid from '@triniti/cms/plugins/dam/components/asset-picker-field/ImageGrid.js';
+import AssetCardGrid from '@triniti/cms/plugins/dam/components/asset-picker-field/AssetCardGrid.js';
+
+export default function getAssetDisplayView(displayView, nodes = [], type) {
+  if (displayView === 'asset-grid') {
```
rsumilang: maybe this should match the component name and use `asset-card-grid`?

rsumilang (changes requested):

src/plugins/dam/utils/getAssetDisplayView.js
guizzettic: Yeah that makes sense to be cleaner

src/plugins/dam/utils/getAssetDisplayView.js
guizzettic: Ok I'll update that here and in cms-core-js repo

rsumilang (approved):

PR #503, opened 2025-10-27
Word count in edit mode

mtakla16:
* Blocksmith :: add live word count in edit mode


src/blocksmith/index.js
```diff
@@ -12,35 +12,43 @@ import DraggableBlockPlugin from '@triniti/cms/blocksmith/plugins/DraggableBlock
...
 import ToolbarPlugin from '@triniti/cms/blocksmith/plugins/ToolbarPlugin.js';
+import WordCountPlugin from '@triniti/cms/blocksmith/plugins/WordCountPlugin.js';
 import HoverInsertButtons from '@triniti/cms/blocksmith/components/hover-insert-buttons/index.js';
 import { useFormContext } from '@triniti/cms/components/index.js';
 import config from '@triniti/cms/blocksmith/config.js';
 
 export default function Blocksmith(props) {
   const { editMode, pbj } = useFormContext();
   const editorRef = useRef(null);
+  const [liveWordCount, setLiveWordCount] = useState(0);
```
efjacobson: @mtakla16 the initial value can come from the node?

src/blocksmith/plugins/WordCountPlugin.js
```diff
@@ -0,0 +1,49 @@
...
+      const root = $getRoot();
+      paragraphWordCount = countWords(root.getTextContent());
+      decoratorWordCount = (root.getChildren().filter(node => $isBlocksmithNode(node))).reduce((acc, curr) => {
+        if (![`${APP_VENDOR}:canvas:block:quote-block`, `${APP_VENDOR}:canvas:block:heading-block`].includes(curr.getCurie())) {
+          return acc;
+        }
+        return acc + countWords(curr.getPbj()?.text || '');
+      }, 0);
+      onWordCountChanged(paragraphWordCount + decoratorWordCount + titleWordCount);
+    });
```
efjacobson, line 24: @mtakla16 this is for the initial number, correct? can you just take that from the node itself?

src/blocksmith/plugins/WordCountPlugin.js
```diff
@@ -0,0 +1,49 @@
+import React, { useEffect } from 'react';
+import { $getRoot } from 'lexical';
+import { useLexicalComposerContext } from '@lexical/react/LexicalComposerContext';
+import { mergeRegister } from '@lexical/utils';
+import { $isBlocksmithNode } from '@triniti/cms/blocksmith/nodes/BlocksmithNode.js';
+import countWords from '@triniti/cms/blocksmith/utils/countWords.js';
```
efjacobson: @mtakla16 is this used anywhere else? if not it can probably just live in this file

src/blocksmith/plugins/WordCountPlugin.js
mtakla16: @efjacobson yes, and no. Both listeners are triggered by [this function](https://github.com/facebook/lexical/blob/814e7d7ff2a9392547f693088e77fbc369a4f643/packages/lexical/src/LexicalUpdates.ts#L505) ([here](https://github.com/facebook/lexical/blob/814e7d7ff2a9392547f693088e77fbc369a4f643/packages/lexical/src/LexicalUpdates.ts#L687) and [here](https://github.com/facebook/lexical/blob/814e7d7ff2a9392547f693088e77fbc369a4f643/packages/lexical/src/LexicalUpdates.ts#L695)) when entering edit mode, but the text content listener has [a wrapper](https://github.com/facebook/lexical/blob/814e7d7ff2a9392547f693088e77fbc369a4f643/packages/lexical/src/LexicalUpdates.ts#L713) that prevents it from firing if the text content has not actually changed. If we passed the initial word count into the plugin, then it would immediately get overwritten with an incorrect number because the decorator listener fires but the text content listener does not. Hence immediately reading the editor state to tally up the initial word count from both text and decorators.

efjacobson: @mtakla16 couldn't you initialize the paragraphWordCount as the word count of the non-decorated nodes from the article?

efjacobson (approved):

PR #509, opened 2025-11-06
Set default stroke_color and stroke_style for divider blocks

guizzettic:
* Blocksmith :: Default values for divider-block fields


joshlevy (approved):
lgtm

PR #512, opened 2025-11-11
Start channel

meghedi:


src/plugins/ovp/components/livestreams-screen/MediaLiveCard.js
```diff
@@ -126,36 +147,40 @@ export default function MediaLiveCard(props) {
             disabled={isRunning ? !canStopChannel : !canStartChannel}
           />
           <ActionButton text="Refresh State" onClick={refresh} color="light" outline disabled={isRefreshing} />
-          <Label className="d-inline">State: {medialive.channelState}</Label>
-          <Icon imgSrc="circle" color={isRunning ? 'danger' : 'dark'} />
+          <span className={className && className.includes('sidebar') ? 'd-block w-100 mt-2' : ''}>
```
efjacobson: @meghedi this is a really weird way to do this...you are applying a class of `d-block w-100 mt-2` when a `className` prop that includes `sidebar` is passed in? couldnt you just do this with css or, failing that, pass in a prop that more clearly indicates that it is in the sidebar?

src/plugins/ovp/components/livestreams-screen/MediaLiveCard.js
```diff
@@ -15,8 +15,24 @@ import getFriendlyErrorMessage from '@triniti/cms/plugins/pbjx/utils/getFriendly
...
+
 export default function MediaLiveCard(props) {
-  const { node, nodeRef, medialive, refresh, isRefreshing } = props;
+  const {
+    node,
+    nodeRef,
+    medialive = defaultMedialive,
+    refresh,
+    isRefreshing = false,
+    showNodeActions = true,
```
efjacobson: @meghedi when i said to reuse a component i didnt mean to reuse the entire conditionally rendered card. you should split this component apart into smaller components, some of which can then be used here (within the card) and also in the sidebar (without the card)

efjacobson (changes requested):

src/plugins/ovp/components/livestreams-screen/MediaLiveCard.js
meghedi: @efjacobson done.

src/plugins/ovp/components/livestreams-screen/MediaLiveCard.js
meghedi: @efjacobson done, I tried moving duplicate code into media-live-channel-controls, but it caused additional duplication elsewhere, so the current version might be the cleanest.

efjacobson (commented):
@meghedi the controls are the only reused thing here, right? if so then do we need separate components for the header and details? what if it the render tree was roughly like this?

```
LivestreamsScreen
  - MediaLiveCard
    - MediaLiveChannelControls
VideoScreen
  - MediaLiveChannelControls
```

src/plugins/ovp/components/video-screen/index.js
```diff
@@ -36,6 +40,30 @@ function VideoScreen(props) {
...
+  const medialiveChannelArn = node.get('medialive_channel_arn');
+  const medialiveRequest = useResolver('triniti:ovp:request:search-videos-request', hasMedialiveChannel && medialiveChannelArn ? {
+    channel: `video-medialive-${nodeRef}`,
+    initialData: {
+      count: 1,
+      page: 1,
+      q: `medialive_channel_arn:"${medialiveChannelArn}"`,
+      derefs: ['medialive_channel_state'],
+    },
+  } : null);
```
efjacobson: @meghedi isn't this included with the get-node-request for the video? iirc it is included by the `HasChannelEnricher`

src/plugins/ovp/components/video-screen/index.js
```diff
@@ -36,6 +41,31 @@ function VideoScreen(props) {
...
+    response: medialiveResponse,
+    run: runMedialiveRequest,
+    isRunning: isRunningMedialiveRequest,
+  } = useRequest(medialiveRequest, Boolean(medialiveRequest));
+
+  const medialiveMetas = medialiveResponse ? medialiveResponse.get('metas', {}) : {};
+
+  const handleRefreshMedialive = () => {
+    runMedialiveRequest();
+    refreshNode();
```
efjacobson: @meghedi why do you need to refresh the video when refreshing the medialive state?

src/plugins/ovp/components/video-screen/index.js
```diff
@@ -36,6 +41,31 @@ function VideoScreen(props) {
...
+    return GetNodeRequestV1.create()
+      .set('node_ref', NodeRef.fromString(`${nodeRef}`))
+      .addToSet('derefs', ['medialive_channel_state']);
+  }, [hasMedialiveChannel, nodeRef]);
+
+  const {
+    response: medialiveResponse,
+    run: runMedialiveRequest,
+    isRunning: isRunningMedialiveRequest,
+  } = useRequest(medialiveRequest, Boolean(medialiveRequest));
```
efjacobson: @meghedi is the second arg here necessary?

scenario 1
- medialiveRequest is null so the second arg is false
- useRequest returns early because second arg is false BUT it would have returned early anyways because the request is null

scenario 2
- medialiveRequest is not null so the second arg is true
- useRequest won't early return now, but the default value of the arg is true anyways

src/plugins/ovp/components/video-screen/index.js
```diff
@@ -36,6 +41,31 @@ function VideoScreen(props) {
...
+  const hasMedialiveChannel = schema.hasMixin('triniti:ovp.medialive:mixin:has-channel') && node.has('medialive_channel_arn');
+
+  const medialiveRequest = React.useMemo(() => {
+    if (!hasMedialiveChannel || !nodeRef) {
+      return null;
+    }
+    return GetNodeRequestV1.create()
+      .set('node_ref', NodeRef.fromString(`${nodeRef}`))
+      .addToSet('derefs', ['medialive_channel_state']);
+  }, [hasMedialiveChannel, nodeRef]);
```
efjacobson, line 54: @meghedi this seems like it should be more of a `useMediaLive` or similar, but i guess since it's only in this component it's alright

src/plugins/ovp/components/video-screen/index.js
meghedi: Done.

src/plugins/ovp/components/video-screen/index.js
meghedi: you are right, we don't need that, runMedialiveRequest() alreadt refreshes

efjacobson (approved):

PR #513, opened 2025-11-12
Duplicate teasers button

mtakla16:
* Add button for duplicating nodes on teasers search screen.


src/plugins/curator/components/search-teasers-screen/index.js
```diff
@@ -128,6 +132,11 @@ function SearchTeasersScreen(props) {
                           <Icon imgSrc="external" alt="open" />
                         </Button>
                       </a>
+                      {canUpdate && (
```
rsumilang: I think you meant `canDuplicate` here

rsumilang:
@mtakla16 might want to check with product if they want to extend this functionality on promotions and widgets page.

rsumilang:
@mtakla16 confirmed with product that this functionality should be added to the all the search pages that the old system had (eg promotions & widgets)

src/plugins/curator/components/useDuplicateNode.js
```diff
@@ -0,0 +1,49 @@
+import startCase from 'lodash-es/startCase.js';
```
rsumilang, line 1: I think this whole file should be moved to a less specific spot since it will also be used for promotions and widgets. Probably a more specific name like useDuplicateNode or something like that which provides the duplication functionality. Maybe wouldn't even care about delegate.

rsumilang (changes requested):

src/plugins/curator/components/search-promotions-screen/index.js
```diff
@@ -20,7 +21,9 @@ function SearchPromotionsScreen(props) {
   const policy = usePolicy();
   const canCreate = policy.isGranted(`${APP_VENDOR}:promotion:create`);
   const canUpdate = policy.isGranted(`${APP_VENDOR}:promotion:update`);
+  const canDuplicate = policy.isGranted(`${APP_VENDOR}:promotion:duplicate`);
```
rsumilang: the correct permission is `create`.... if you can create, you can duplicate.

src/plugins/curator/components/search-widgets-screen/index.js
```diff
@@ -78,7 +80,9 @@ function SearchWidgetsScreen(props) {
               {response.get('nodes', []).map(node => {
                 const ref = node.generateNodeRef();
                 const canUpdate = policy.isGranted(`${ref.getQName()}:update`);
+                const canDuplicate = policy.isGranted(`${ref.getQName()}:duplicate`);
```
rsumilang: the correct permission is `create`.... if you can create, you can duplicate.

rsumilang (changes requested):
looks good. minor changes.

rsumilang (approved):

PR #514, opened 2025-11-13
multiselect

rsumilang:


src/plugins/ncr/components/node-picker-field/MultiSelectField.js
```diff
@@ -36,19 +38,109 @@ export default function MultiSelectField(props) {
...
+    if (action.action === 'input-change') {
+      request.set('q', value);
+      setQ(value);
+      if (value) setCachedQuery(value);
+      return;
+    }
+    if (action.action === 'menu-close') {
+      setMenuIsOpen(false);
+      clearQuery();
+    }
```
guizzettic: Can we just clear when we call `onMenuClose` below, this way we avoid multiple close attempts?

src/plugins/ncr/components/node-picker-field/MultiSelectField.js
```diff
@@ -36,19 +38,109 @@ export default function MultiSelectField(props) {
...
+  useEffect(() => {
+    const handleKeyDown = (e) => {
+      if (isModifierPressed(e)) modifierKeyPressedRef.current = true;
+    };
+    const handleKeyUp = (e) => {
+      if (!isModifierPressed(e)) modifierKeyPressedRef.current = false;
+    };
+    const handleMouseDown = (e) => {
+      if (isModifierPressed(e)) modifierKeyPressedRef.current = true;
+    };
```
guizzettic, line 74: I'm seeing multiselect for one more image once cmd key is let go, instead of reset immediately. Not sure if that's fine for this

rsumilang: @guizzettic yes the intent is to keep showing the same list if the modifier key was held down. once the person lets go then it wont used the cached results again. however if the user wanted a different result they can still type something new or just click out of the box.

guizzettic (approved):
Nice looks good 👍

PR #515, opened 2025-11-17
History card user

meghedi:


meghedi:
@efjacobson thank you for clarifying the issue, I was able to resolve it, can you please review. thanks

efjacobson (approved):

PR #519, opened 2025-11-19
Screen :: Adding support for a component in tabs.

rsumilang:
* Screen :: Adding support for a component in tabs.

### DIFF
--- a/src/components/screen/index.js
+++ b/src/components/screen/index.js
@@ -130,10 +130,26 @@ export default function Screen(props) {
 
         {tabs.length > 0 && (
           <Nav underline className="screen-navtabs">
-            {tabs.map((tab) => {
+            {tabs.map((tab, index) => {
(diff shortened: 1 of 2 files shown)


src/components/screen/index.js
```diff
@@ -130,10 +130,26 @@ export default function Screen(props) {
...
+              if (tab.component) {
+                const Component = tab.component;
+                return (
+                  <Component
+                    key={index}
+                    text={tab.text}
+                    to={tab.to}
+                    activeTab={activeTab}
+                    {...(tab.componentProps || {})}
+                  />
```
efjacobson, line 151: @rsumilang is ther a reason not to continue using `tab.to` for the key? fyi react falls back to using the index when there is no key

also - does it actually need to know what the active tab is, or only if it is active?

efjacobson (approved):

PR #521, opened 2025-11-20
Multiselect

rsumilang:


src/plugins/ncr/components/node-picker-field/MultiSelectField.js
```diff
@@ -36,25 +37,71 @@ export default function MultiSelectField(props) {
...
+  } = useAsyncPaginate({
+    loadOptions,
+    debounceTimeout,
+    additional: { page: 1, request },
+    filterOption: null,
+  });
+
+  const handleMenuClose = useCallback(() => {
+    setSearchValue('');
+  }, []);
```
efjacobson: @rsumilang what does `useCallback` do for you here? seems like not much since `onBlur` is a new fn every render

src/plugins/ncr/components/node-picker-field/MultiSelectField.js
```diff
@@ -36,25 +37,71 @@ export default function MultiSelectField(props) {
...
+    return value;
+  }, [searchValue]);
+
+  const handleAsyncInputChange = useCallback((value, meta) => {
+    const nextValue = handleInputChange(value, meta);
+    if (typeof nextValue !== 'undefined') {
+      return asyncOnInputChange?.(nextValue, meta);
+    }
+    return asyncOnInputChange?.(value, meta);
+  }, [handleInputChange, asyncOnInputChange]);
```
efjacobson: @rsumilang this seems odd too - i think using `handleInputChange` as a dependency here is just an extra step as `searchValue` is the real dependency

how you have it set up now:

```javascript
const handleInputChange = useCallback(() => { ... }, [searchValue]); // handleInputChange will only change when searchValue changes
const handleAsyncInputChange = useCallback(() => { ... }, [handleInputChange, asyncOnInputChange]); // handleAsyncInputChange will only change when handleInputChange or asyncOnInputChange changes
```

but this would do the same thing (you just end up putting the `handleInputChange` logic inside `handleAsyncInputChange` rather than caching two functions:

```javascript
const handleAsyncInputChange = useCallback(() => { ... }, [searchValue, asyncOnInputChange]);
```

something like this, i think:

```javascript
  const handleAsyncInputChange = useCallback((value, meta) => {
    let nextValue;
    switch (meta.action) {
      case 'input-change':
        nextValue = value;
        setSearchValue(nextValue);
        break;
      case 'menu-close':
      case 'input-blur':
        nextValue = '';
        setSearchValue(nextValue);
        break;
      case 'set-value':
        nextValue = searchValue;
        break;
      default:
        nextValue = value;
    }
    return asyncOnInputChange?.(nextValue, meta);
  }, [searchValue, asyncOnInputChange]);
```

efjacobson: that said, is the purpose of useCallback here to prevent the component from re-rendering when a selection is made?

rsumilang: @efjacobson The useCallback here just memoizes the function so the child components don't see a new function handler every render. It will still re-render for each selection made.

In regards to handleInputChange, yes it can be combined with handleAsyncInputChange if you prefer. I think keeping them separate was easier to understand.

efjacobson: > @efjacobson The useCallback here just memoizes the function so the child components don't see a new function handler every render.

- yes, but why? what is accomplished by making sure the child components don't see a new function handler every render?

> In regards to handleInputChange, yes it can be combined with handleAsyncInputChange if you prefer. I think keeping them separate was easier to understand.

- its at least a little odd. along with obfuscating the dependency on `searchValue`, `handleInputChange` is kind of a confusing name - it's not handling the input any more than `handleAsyncInputChange`, nor is the input itself somehow sync or async. what `handleInputChange` really does is set the `searchValue`

rsumilang: it keeps the same handler instance flowing into ReactSelect, so its memoized input/menu pieces and debounced loader dont restart every render. I can yank it if you'd rather keep things simpler, but this just helps the async picker stay snappy.

CHANGELOG-1.x.md
```diff
@@ -2,6 +2,7 @@
 
 
 ## vN.N.N
+* NCR MultiSelectField :: Adding ability to mutli select options while holding modifier key.
```
efjacobson: @rsumilang this is no longer accurate

CHANGELOG-1.x.md
```diff
@@ -2,7 +2,7 @@
 
 
 ## vN.N.N
-* NCR MultiSelectField :: Adding ability to mutli select options while holding modifier key.
+* NCR MultiSelectField :: Adding ability to mutli select options.
```
efjacobson: @rsumilang typo + this is not terribly helpful as we already can multi select options, that's not what you're adding here

efjacobson (approved):

rsumilang:
@efjacobson updated the changelog after you approved. let me know if still clear.

efjacobson (approved):

PR #529, opened 2026-01-15
Add insert text block

tbone1020:


src/blocksmith/index.js
```diff
@@ -11,6 +11,7 @@ import BlocksmithPlugin from '@triniti/cms/blocksmith/plugins/BlocksmithPlugin.j
 import DraggableBlockPlugin from '@triniti/cms/blocksmith/plugins/DraggableBlockPlugin.js';
 import FloatingTextFormatToolbarPlugin from '@triniti/cms/blocksmith/plugins/FloatingTextFormatToolbarPlugin.js';
 import InsertBlockPlugin from '@triniti/cms/blocksmith/plugins/InsertBlockPlugin.js';
+import PersistentInsertTextBlockPlugin from '@triniti/cms/blocksmith/plugins/PersistentInsertTextBlockPlugin.js';
```
efjacobson: @tbone1020 "persistent" is kind of a weird name to use for this. the one at the top is also persistent

efjacobson (changes requested):
@tbone1020 
- the button should not be visible when in view mode
- please style it the same as the other green text block button (the floating one, like we discussed the other day)
- it looks odd because it is centered but the floating buttons are not (they are 1rem less than full width). either center the floating buttons or do the same 1rem shift for this one

tbone1020:
@efjacobson ready for review again.

src/blocksmith/index.js
tbone1020: done

efjacobson (approved):

PR #534, opened 2026-02-27
Apply slug normalization on blur

rsumilang:


src/plugins/ncr/components/slug-field/RenameForm.js
```diff
@@ -98,9 +98,8 @@ function RenameForm(props) {
             name="slug"
             label="New Slug"
             required
-            format={value => value?.trim()}
+            format={withDatedSlug ? parseDatedSlug : parseSlug}
```
guizzettic: I noticed that we aren't trimming trailing spaces/dashes/slashes anymore, causing validation errors if left.

guizzettic (changes requested):
I think we should resolve the trailing items issue (we aren't trimming trailing spaces/dashes/slashes anymore, causing validation errors and manual cleanup in order to submit).

I also noticed that the slug when creating an article still jumps ahead, seems tangentially related to this ticket. But I don't think that's an issue since it tracks the article title... but just wanted to mention it in case you think it should be resolved there as well.

guizzettic:
@rsumilang Is this ready for another PR review?

rsumilang:
@guizzettic yes please

guizzettic (changes requested):
@rsumilang 
1) Create article still has a stale slug issue; after teh initial title blur creates the slug, later title changes do not refresh it unless user manually fixes the slug. I think it should continue syncing with title updates unless overridden
2) [This expected behavior](https://github.com/tmz-apps/tmz-cms/issues/1563#:~:text=issue%201%20%2D%20when%20only%20a%20title%20has%20been%20entered%20(no%20slug)%2C%20the%20button%20should%20be%20active.%20when%20clicked%2C%20the%20title%20should%20be%20sanitized%20as%20normal%20and%20the%20slug%20should%20be%20auto%2Dgenerated) for create article modal as well. It sounds like the create article button should be active when a title has been entered without having to click outside of title field.

rsumilang:
@guizzettic thanks, nice finds. please check now.

guizzettic:
@rsumilang Looks better, still seeing an Enter submit issue in the Create Article modal. 

Repro:
- enter title: `slug test`
- blur title field so slug auto generates
- go back and update title: `slug test 2`
- press Enter to submit

Result: submit fails with `Code: ALREADY_EXISTS`

Looks as though the Enter path is still submitting with the previously auto generated slug, instead of updated slug. Clicking Create Article button appears to work, so this seems specific to Enter submit workflow.

image

rsumilang:
> @rsumilang Looks better, still seeing an Enter submit issue in the Create Article modal.
> 
> Repro:
> 
> * enter title: `slug test`
> * blur title field so slug auto generates
> * go back and update title: `slug test 2`
> * press Enter to submit
> 
> Result: submit fails with `Code: ALREADY_EXISTS`
> 
> Looks as though the Enter path is still submitting with the previously auto generated slug, instead of updated slug. Clicking Create Article button appears to work, so this seems specific to Enter submit workflow.
> 
> image

I think you got that error because you created a similar slug too fast. Try again.

guizzettic:
@rsumilang 
Yeah that was not a clear example, I retested it with unique slugs and found the same issue. Ultimately this may not be a concern since the collision of two articles having the same slug isn't high but still points to Enter submit path using original auto generated slug, even though form UI has updated.

To actually repro:
1) Create a new article with unique title (in my case i did `alpha brick`) and submit
2) Create new article again and set title initially to previously created unique title 
3) Leave the title field
4) Update title to something new and completely unique (in my case i updated it to `zebra planet`) and press Enter.

Article is not created due to collision with the initial `alpha brick` slug even though it was updated.

src/plugins/news/components/create-article-modal/index.js
guizzettic, line 70: @rsumilang Pressing Enter submits here, but the slug resync only happens in `handleTitleBlur`. In the repro I shared, this path still appears to submit the original slug.

src/plugins/news/components/create-article-modal/index.js
```diff
@@ -45,13 +31,14 @@ function CreateArticleModal(props) {
   delegate.handleSubmit = async (values) => {
     try {
       await progressIndicator.show('Creating Article...');
-      if (values.slug && isValidDatedSlug(values.slug)) {
-        values.slug = values.slug.toLowerCase();
-      } else if (values.slug && !isValidDatedSlug(values.slug)) {
-        values.slug = addDateToSlug(values.slug).toLowerCase();
+      const rawSlug = (values.slug ?? '').trim();
+      const rawTitle = (values.title ?? '').trim();
+      if (rawSlug) {
```
guizzettic: @rsumilang This seems to treat non empty rawSlug as authoritative on submit, seems to be why the Enter path can preserve the original auto generated slug instead of regenerating from latest title.

rsumilang:
> @rsumilang Yeah that was not a clear example, I retested it with unique slugs and found the same issue. Ultimately this may not be a concern since the collision of two articles having the same slug isn't high but still points to Enter submit path using original auto generated slug, even though form UI has updated.
> 
> To actually repro:
> 
> 1. Create a new article with unique title (in my case i did `alpha brick`) and submit
> 2. Create new article again and set title initially to previously created unique title
> 3. Leave the title field
> 4. Update title to something new and completely unique (in my case i updated it to `zebra planet`) and press Enter.
> 
> Article is not created due to collision with the initial `alpha brick` slug even though it was updated.

I think you found a weird bug that the desk was mentioning last week. Try to break it now.

guizzettic:
> > @rsumilang Yeah that was not a clear example, I retested it with unique slugs and found the same issue. Ultimately this may not be a concern since the collision of two articles having the same slug isn't high but still points to Enter submit path using original auto generated slug, even though form UI has updated.
> > To actually repro:
> > 
> > 1. Create a new article with unique title (in my case i did `alpha brick`) and submit
> > 2. Create new article again and set title initially to previously created unique title
> > 3. Leave the title field
> > 4. Update title to something new and completely unique (in my case i updated it to `zebra planet`) and press Enter.
> > 
> > Article is not created due to collision with the initial `alpha brick` slug even though it was updated.
> 
> I think you found a weird bug that the desk was mentioning last week. Try to break it now.

Did you try to repro? Same issue happening, I believe the handleKeyDown needs to be updated as well.

rsumilang:
> > > @rsumilang Yeah that was not a clear example, I retested it with unique slugs and found the same issue. Ultimately this may not be a concern since the collision of two articles having the same slug isn't high but still points to Enter submit path using original auto generated slug, even though form UI has updated.
> > > To actually repro:
> > > 
> > > 1. Create a new article with unique title (in my case i did `alpha brick`) and submit
> > > 2. Create new article again and set title initially to previously created unique title
> > > 3. Leave the title field
> > > 4. Update title to something new and completely unique (in my case i updated it to `zebra planet`) and press Enter.
> > > 
> > > Article is not created due to collision with the initial `alpha brick` slug even though it was updated.
> > 
> > 
> > I think you found a weird bug that the desk was mentioning last week. Try to break it now.
> 
> Did you try to repro? Same issue happening, I believe the handleKeyDown needs to be updated as well.

3rd time's a charm.

guizzettic (approved):
Approved, conditional on updating tickets with PR link and proper QA notes 👍

PR #535, opened 2026-03-07
Linked assets sizes

tbone1020:


src/plugins/dam/components/linked-assets-card/index.js
```diff
@@ -76,7 +96,17 @@ function LinkedAssetsCard(props) {
...
       <Card>
         <CardHeader>
-          <span>Linked Assets {isRunning && <Spinner />}</span>
+          <span style={{position: 'relative'}}>
+            Linked Assets {isRunning && <Spinner style={{position: 'absolute', top: '5px'}} />}
+          </span>
+          <ResizeGallerySlider 
+            imagesPerRow={imagesPerRow}
+            maxImagesPerRow={MAX_IMAGES_PER_ROW}
+            minIMagesPerRpw={MIN_IMAGES_PER_ROW}
```
rsumilang: typo?

src/plugins/dam/components/linked-assets-card/index.js
```diff
@@ -68,6 +72,22 @@ function LinkedAssetsCard(props) {
...
   };
+  
+  const handleIncreaseImagesPerRow = () => {
+    if (imagesPerRow <= MAX_IMAGES_PER_ROW) {
+      setImagesPerRow(imagesPerRow + 1);
+    }
+  };
+
+  const handleDecreaseImagesPerRow = () => {
+    if (imagesPerRow >= MIN_IMAGES_PER_ROW) {
```
rsumilang: Prefer if (imagesPerRow > MIN_IMAGES_PER_ROW) so it never goes below 1.

src/plugins/dam/components/linked-assets-card/index.js
```diff
@@ -43,6 +44,9 @@ function LinkedAssetsCard(props) {
   const canLink = policy.isGranted('triniti:dam:command:link-assets');
   const canUnlink = policy.isGranted('triniti:dam:command:unlink-assets');
   const batch = useBatch(response);
+  const MAX_IMAGES_PER_ROW = 12;
+  const MIN_IMAGES_PER_ROW = 1;
+  const [ imagesPerRow, setImagesPerRow ] = useState(7);
```
rsumilang: useState(7) for initial imagesPerRow is fine. an optional comment or DEFAULT_IMAGES_PER_ROW = 7 would make intent clearer.

rsumilang (changes requested):

src/plugins/dam/components/linked-assets-card/index.js
tbone1020: Yes this is a typo. Fixed.

src/plugins/dam/components/linked-assets-card/index.js
tbone1020: changed

src/plugins/dam/components/linked-assets-card/index.js
tbone1020: Done

src/plugins/dam/components/linked-assets-card/index.js
```diff
@@ -68,6 +73,22 @@ function LinkedAssetsCard(props) {
     // note that the modal would have already done the linking
     run();
   };
+  
+  const handleIncreaseImagesPerRow = () => {
+    if (imagesPerRow <= MAX_IMAGES_PER_ROW) {
```
rsumilang: The <= guard let it increment to 13 when the max is 12.

src/plugins/dam/components/asset-picker-field/AssetCardGrid.js
```diff
@@ -33,8 +32,13 @@ function AssetCard ({ node, batch, onSelectAsset }) {
     }
   };
 
+  const Component = imagesPerRow ? "div" : Col;
+  const layout = imagesPerRow
+    ? { style: { width: `calc(100% / ${imagesPerRow})` } }
+    : { xs:12, sm: 6, md: 4, lg: 3, xl: "2p" }
```
rsumilang: Lint.

rsumilang (changes requested):

src/plugins/dam/components/linked-assets-card/index.js
tbone1020: Done

src/plugins/dam/components/asset-picker-field/AssetCardGrid.js
tbone1020: Done

rsumilang:
@tbone1020 Seems like some requirements are not met yet.

1. assets tab: slider does nothing.
2. add image to gallery modal: images still cropped.
3. add image to gallery modal: i see the title overlay and they seem to be all different sizes depending on the text. can we give it a fixed height for 2 lines of text and if it exceeds 2 lines of text then we can just truncate it with "...". i believe this can be done with css.

i see some other changes not done but i suppose they are going to be done in core?

tbone1020:
@rsumilang The asset tab is uses the [AssetCardGrid Component](https://github.com/triniti/cms-js/blob/linked-assets-sizes/src/plugins/dam/components/asset-picker-field/AssetPresenter.js) which is set at the vendor level.

I should have mentioned this but you will have to either change `AssetTableLinked` in the assetPresenter to `AssetCardGrid` or add "displayView='asset-card-grid'" to the `LinkedAssetsCard` component.

I'll take a look at the other issues though.

rsumilang:
@tbone1020 my initial code review seems to be true still.

tbone1020:
@rsumilang should be fixed now. 
I wasn't able to get 2 lines to work so I kept it to 1 line and then an ellipsis.

rsumilang:
@tbone1020 

1. asset tab: the slider still does nothing when its an asset table
2. asset tab: the slider bounces around when i select an item in the grid when the delete button shows up
3. the text truncation shouldn't be set to a hard coded number. it should display as much as possible and truncate the rest.
4. the default view for the grid seems very large. does that match the sizing of the grid that we currently have?

tbone1020:
@rsumilang 
1. As we discussed earlier, the slider will start working once you make the layout code change.
2. I positioned the slider in the asset and gallery image tab so that the delete button shouldn't affect the layout regardless if it exists or not.
3. I removed the hard coded width. The truncation should be more dynamic now.
4. Do you mean the width of the container? They want it to match the width of the gallery images container.

rsumilang:
@tbone1020 

1. But why render a slider that does nothing to the asset table? 
2. The slider doesn't seem like it's in the center of the title bar. It seems kind of off to the left in a weird position. I think the placement should be dead on center if we're keeping it in the title bar.
3. That looks better. After speaking to product, they'd like to show as much as possible when mousing over the text as well. So on mouse over of an item you can just remove the "nowrap" css which should be sufficient.
4. I don't mean the width of the container, I mean the width of the items. They start off on my computer supersized. It seems that they change sizes just by resizing the window. I think the size that they start off with should be the same size that they are in the linked assets view. 

CleanShot 2026-04-08 at 17 17 40

rsumilang:
@tbone1020 Expanding on item 4, it should follow the exact same thumb treatment that they get in the modal when resizing the window. Basically, use the same breakpoint styling. You can see that they are different when you resize the page smaller and bigger. 

Here are some examples.

CleanShot 2026-04-21 at 15 36 21
CleanShot 2026-04-21 at 15 36 49
CleanShot 2026-04-21 at 15 37 10

efjacobson:
@rsumilang there's no expectation that the slider would affect the modal. the current implementation on the gallery images tab doesn't do that (intentionally)

rsumilang:
> @rsumilang there's no expectation that the slider would affect the modal. the current implementation on the gallery images tab doesn't do that (intentionally)

I don't have that expectation either. I'm merely saying the starting size of the thumbnail should match the styling of the modal. I'm not saying adjusting the slider should adjust the images in the modal. 

edit: i see the gallery image tab is different and this matches that.

rsumilang (approved):

PR #537, opened 2026-04-08
Add seo alternate urls to People page

guizzettic:
* Add support for seo_alternate_urls field on People SEO tab.

### DIFF
--- a/CHANGELOG-1.x.md
+++ b/CHANGELOG-1.x.md
@@ -54,6 +54,7 @@
 * Blocksmith :: adjust drag image position when dragging blocks in the editor
 * asset linker modal should default to image assets only
 * Display `is_vertical` field on video asset details tab and video mezzanine card.
+* Add seo_alternate_urls field for People pages.
 
 
 ## v1.2.2
--- a/CHANGELOG-1.x.md
+++ b/CHANGELOG-1.x.md
@@ -1,4 +1,4 @@
-# CHANGELOG
+****# CHANGELOG
 
 
 ## vN.N.N
@@ -55,7 +55,7 @@
 * asset linker modal should default to image assets only
 * Display `is_vertical` field on video asset details tab and video mezzanine card.
 * Customizable keyValuesField validateKeys.
-* Add support for seo_alternate_urls field on People SEO tab.
+* Add SeoAlternateUrlsField to SEO tab.
 
 
 ## v1.2.2
--- a/CHANGELOG-1.x.md
+++ b/CHANGELOG-1.x.md
@@ -1,4 +1,4 @@
-****# CHANGELOG
+# CHANGELOG
 
 
 ## vN.N.N
--- a/CHANGELOG-1.x.md
+++ b/CHANGELOG-1.x.md
@@ -54,7 +54,7 @@
 * Blocksmith :: adjust drag image position when dragging blocks in the editor
 * asset linker modal should default to image assets only
 * Display `is_vertical` field on video asset details tab and video mezzanine card.
-* Add seo_alternate_urls field for People pages.
+* Add support for seo_alternate_urls field on People SEO tab.
 
 
 ## v1.2.2
--- a/src/plugins/common/components/seo-alternate-urls-field/index.js
+++ b/src/plugins/common/components/seo-alternate-urls-field/index.js
@@ -1,0 +1,76 @@
+import React from 'react';
+import { FormText } from 'reactstrap';
+import { FlatArrayField, UrlField, useFormContext } from '@triniti/cms/components/index.js';
+
+const normalizeSeoAlternateUrl = value => `${value || ''}`.trim();
+
+const supportsSeoAlternateUrls = (schema) => {
+  if (!schema || typeof schema.hasField !== 'function') {
+    return false;
+  }
+
+  return schema.hasField('seo_alternate_urls');
+};
+
+const getNormalizedUrls = (allValues, fieldName) => {
+  const values = Array.isArray(allValues?.[fieldName]) ? allValues[fieldName] : [];
+  return values
+    .map(normalizeSeoAlternateUrl)
+    .filter(Boolean)
+    .map(value => value.toLowerCase());
+};
+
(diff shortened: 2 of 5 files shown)


src/plugins/people/components/person-screen/index.js
```diff
@@ -108,7 +108,7 @@ function PersonScreen(props) {
             <DetailsTab {...props} />
           </TabPane>
           <TabPane tabId="seo">
-            <SeoTab {...props} />
+            <SeoTab {...props} showAlternateUrls />
```
efjacobson: @guizzettic is the prop necessary? seems like the schema check would prevent the form field from rendering on nodes that don't have the schema field

src/plugins/common/components/seo-tab/index.js
```diff
@@ -1,10 +1,24 @@
...
+
+const seoAlternateUrlValidator = (value) => {
+  const normalized = normalizeSeoAlternateUrl(value);
+  if (!normalized) {
+    return 'Required';
+  }
+
+  if (!normalized.startsWith('https://')) {
+    return 'Must start with https://.';
+  }
```
efjacobson: @guizzettic cleanest way to do this is probably 

```javascript
 if (new URL(normalized).protocol !== 'https:') {
```

src/plugins/common/components/seo-tab/index.js
```diff
@@ -40,6 +58,21 @@ export default function SeoTab() {
...
         />
         <SelectField name="meta_keywords" label="Meta Keywords" allowOther isMulti />
+        {showSeoAlternateUrls && (
+          <>
+            <FlatArrayField
+              name="seo_alternate_urls"
+              label="Alternate URLs"
+              component={UrlField}
+              parse={normalizeSeoAlternateUrl}
+              format={normalizeSeoAlternateUrl}
```
efjacobson: @guizzettic both parse and format necessary here?

efjacobson (changes requested):

src/plugins/common/components/seo-tab/index.js
guizzettic: @efjacobson No that was a misunderstanding, we don't need the format.

src/plugins/common/components/seo-tab/index.js
```diff
@@ -40,9 +44,10 @@ function DescriptionWarning(props) {
 }
 
 export default function SeoTab(props) {
-  const { node, showAlternateUrls = false } = props;
+  const { node } = props;
   const schema = node.schema();
-  const showSeoAlternateUrls = showAlternateUrls && schema.hasField('seo_alternate_urls');
+  const showSeoAlternateUrls = schema.hasMixin('triniti:people:mixin:person')
+    && schema.hasField('seo_alternate_urls');
```
efjacobson: @guizzettic do you need to check that it is a person?

guizzettic: @efjacobson Yes I believe we do since seo_alternate_urls  lives on `triniti:common:mixing:seo`. Scope says people pages, so I do the person specific check

efjacobson: @guizzettic true but we shouldn't make assumptions like that at the triniti level. i think the best way forward with this is to make it a dedicated Field (similar to SeoTitleField) so that it
1. encompasses all of its own logic (and at this level, only checks for field availability, not node type)
2. is more convenient to override at the vendor level

guizzettic: @efjacobson Yeah you're right, I had misunderstood. I refactored to have its own field and then at vendor level we can override

src/plugins/common/components/seo-alternate-urls-field/index.js
```diff
@@ -0,0 +1,76 @@
...
+
+  return (
+    <>
+      <FlatArrayField
+        {...rest}
+        name={name}
+        label={label}
+        component={UrlField}
+        parse={normalizeSeoAlternateUrl}
+        validator={createSeoAlternateUrlValidator(name)}
```
efjacobson: @guizzettic is it necessary to create a new validator for every render?

CHANGELOG-1.x.md
```diff
@@ -1,4 +1,4 @@
-# CHANGELOG
+****# CHANGELOG
```
efjacobson: @guizzettic oops

src/plugins/common/components/seo-alternate-urls-field/index.js
guizzettic: @efjacobson No it's not, it was to close over the field name for duplicate checking. I cleaned it up

src/plugins/common/components/seo-alternate-urls-field/index.js
```diff
@@ -0,0 +1,68 @@
...
+
+  try {
+    if (new URL(normalized).protocol !== 'https:') {
+      return 'Must use https://.';
+    }
+  } catch {
+    return 'Must be a valid URL.';
+  }
+
+  const baseName = meta?.name?.replace(/\[\d+]$/, '') || 'seo_alternate_urls';
```
efjacobson: @guizzettic why even attempt to get it from meta? this function is only used for this purpose in this component

guizzettic: @efjacobson Yeah this is unnecessary after validator factory was removed.

src/plugins/common/components/seo-alternate-urls-field/index.js
```diff
@@ -0,0 +1,61 @@
...
+  const { pbj } = useFormContext();
+
+  if (!pbj.schema().hasField('seo_alternate_urls')) {
+    return null;
+  }
+
+  return (
+    <>
+      <FlatArrayField
+        {...props}
```
efjacobson, line 49: @guizzettic don't think it takes any props but not causing any harm i suppose

efjacobson (approved):

PR #539, opened 2026-04-14
View launch text

guizzettic:
* Pass gallery launch text through to poster image preview.

### DIFF
--- a/src/blocksmith/components/gallery-block-modal/index.js
+++ b/src/blocksmith/components/gallery-block-modal/index.js
@@ -4,12 +4,16 @@ import { SwitchField, TextField } from '@triniti/cms/components/index.js';
 import AspectRatioField from '@triniti/cms/plugins/common/components/aspect-ratio-field/index.js';
 import GalleryPickerField from '@triniti/cms/blocksmith/components/gallery-block-modal/GalleryPickerField.js';
 import ImageAssetPickerField from '@triniti/cms/plugins/dam/components/image-asset-picker-field/index.js';
+import useNode from '@triniti/cms/plugins/ncr/components/useNode.js';
 import withBlockModal from '@triniti/cms/blocksmith/components/with-block-modal/index.js';
 import AsideField from '@triniti/cms/blocksmith/components/with-block-modal/AsideField.js';
 
 function GalleryBlockModal(props) {
   const { nodeRef: containerRef } = props.containerFormContext;
   const { values } = useFormState({ subscription: { values: true } });
+  const { node: gallery } = useNode(values.node_ref || null);
+  const launchText = values.launch_text || gallery?.get('launch_text') || '';
+
   return (
     <>
       <GalleryPickerField name="node_ref" label="Gallery" required />
@@ -19,6 +23,7 @@ function GalleryBlockModal(props) {
         description="When not set, the gallery's image will be used."
         nodeRef={containerRef}
         galleryRef={values.node_ref}
+        previewProps={{ launchText }}
       />
       <SwitchField
         name="start_at_poster"
--- a/src/blocksmith/components/gallery-block-modal/index.js
+++ b/src/blocksmith/components/gallery-block-modal/index.js
@@ -10,9 +10,9 @@ import AsideField from '@triniti/cms/blocksmith/components/with-block-modal/Asid
 
 function GalleryBlockModal(props) {
   const { nodeRef: containerRef } = props.containerFormContext;
-  const { values } = useFormState({ subscription: { values: true } });
-  const { node: gallery } = useNode(values.node_ref || null);
-  const launchText = values.launch_text || gallery?.get('launch_text') || '';
+  const { values = {} } = useFormState({ subscription: { values: true } });
+  const { node: gallery } = useNode(values.node_ref);
+  const launchText = values.launch_text || gallery?.get('launch_text', '') || '';
 
   return (
     <>
--- a/src/plugins/dam/components/image-asset-picker-field/index.js
+++ b/src/plugins/dam/components/image-asset-picker-field/index.js
@@ -4,14 +4,19 @@ import AssetPickerField from '@triniti/cms/plugins/dam/components/asset-picker-f
 import damUrl from '@triniti/cms/plugins/dam/damUrl.js';
 
 function Preview(props) {
-  const { assetRef } = props;
+  const { assetRef, launchText = '' } = props;
   const downloadUrl = damUrl(assetRef);
   const previewUrl = damUrl(assetRef, '1by1', 'sm');
 
   return (
     <div className="d-block mb-3">
-      <a href={downloadUrl} target="_blank" rel="noopener noreferrer" className="hover-box-shadow d-inline-block rounded-3">
+      <a href={downloadUrl} target="_blank" rel="noopener noreferrer" className="hover-box-shadow d-inline-block rounded-3 overflow-hidden position-relative">
         <Media src={previewUrl} alt="" width={200} height={200} object className="rounded-3" />
+        {launchText && (
+          <div className="position-absolute bottom-0 start-0 end-0 px-2 py-1 text-white text-break bg-black bg-opacity-75 lh-sm">
+            {launchText}
+          </div>
+        )}
       </a>
     </div>
   );
--- a/src/plugins/dam/components/image-asset-picker-field/index.js
+++ b/src/plugins/dam/components/image-asset-picker-field/index.js
@@ -13,7 +13,7 @@ function Preview(props) {
       <a href={downloadUrl} target="_blank" rel="noopener noreferrer" className="hover-box-shadow d-inline-block rounded-3 overflow-hidden position-relative">
         <Media src={previewUrl} alt="" width={200} height={200} object className="rounded-3" />
         {launchText && (
-          <div className="position-absolute bottom-0 start-0 end-0 px-2 py-1 text-white text-break bg-black bg-opacity-75 lh-sm">
+          <div className="position-absolute bottom-0 start-0 end-0 px-2 py-1 text-white text-break text-uppercase bg-black bg-opacity-75 lh-sm">
             {launchText}
           </div>
         )}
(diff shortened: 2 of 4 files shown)


src/blocksmith/components/gallery-block-modal/index.js
```diff
@@ -4,12 +4,16 @@ import { SwitchField, TextField } from '@triniti/cms/components/index.js';
...
 import GalleryPickerField from '@triniti/cms/blocksmith/components/gallery-block-modal/GalleryPickerField.js';
 import ImageAssetPickerField from '@triniti/cms/plugins/dam/components/image-asset-picker-field/index.js';
+import useNode from '@triniti/cms/plugins/ncr/components/useNode.js';
 import withBlockModal from '@triniti/cms/blocksmith/components/with-block-modal/index.js';
 import AsideField from '@triniti/cms/blocksmith/components/with-block-modal/AsideField.js';
 
 function GalleryBlockModal(props) {
   const { nodeRef: containerRef } = props.containerFormContext;
   const { values } = useFormState({ subscription: { values: true } });
+  const { node: gallery } = useNode(values.node_ref || null);
```
bhsiao55: @guizzettic 
  const { node: gallery } = useNode(values.node_ref || null);
do you really need the || null here?  Are you protecting this in case values is undefined?  Cause in ImageAssetPickerField we are using the same thing for galleryRef prop. Maybe if we did this instead
 const { values = {} } = useFormState({ subscription: { values: true } });
everything might be cleaner

src/blocksmith/components/gallery-block-modal/index.js
```diff
@@ -4,12 +4,16 @@ import { SwitchField, TextField } from '@triniti/cms/components/index.js';
...
+  const launchText = values.launch_text || gallery?.get('launch_text') || '';
```
bhsiao55: @guizzettic probably need a default if launch_text doesn't exist on the gallery node
 const launchText = values.launch_text || gallery?.get('launch_text', '') || '';

bhsiao55 (changes requested):

src/blocksmith/components/gallery-block-modal/index.js
guizzettic: @bhsiao55  Good call, updated 👍

bhsiao55:
@guizzettic I see you are adding the launchtext at the triniti level but you are not really rendering the text here.  I assume you will be rendering it in cms-core but just wondering why you decided to add the launchtext here and not just do everything in cms-core?

guizzettic:
> @guizzettic I see you are adding the launchtext at the triniti level but you are not really rendering the text here. I assume you will be rendering it in cms-core but just wondering why you decided to add the launchtext here and not just do everything in cms-core?

@bhsiao55 The thinking is that the gallery block is owned by triniti and cms-core doensn't override it, so the clean place to source the selected galleries `launch_text` and pass to poster preview is shared triniti layer. Then rendering would still live in cms-core and we wouldn't have to duplicate the gallery block modal to pass one prop.

bhsiao55:
> > @guizzettic I see you are adding the launchtext at the triniti level but you are not really rendering the text here. I assume you will be rendering it in cms-core but just wondering why you decided to add the launchtext here and not just do everything in cms-core?
> 
> @bhsiao55 The thinking is that the gallery block is owned by triniti and cms-core doensn't override it, so the clean place to source the selected galleries `launch_text` and pass to poster preview is shared triniti layer. Then rendering would still live in cms-core and we wouldn't have to duplicate the gallery block modal to pass one prop.

@guizzettic you can override that gallery block in cms-core as well.  We probably don't have that file in cms-core cause there was no need to change anything triniti was already doing.  To me it feels weird to be adding something that's not going to be used. If we are going to be adding this prop in triniti, I would render it in triniti as well

guizzettic:
> > > @guizzettic I see you are adding the launchtext at the triniti level but you are not really rendering the text here. I assume you will be rendering it in cms-core but just wondering why you decided to add the launchtext here and not just do everything in cms-core?
> > 
> > 
> > @bhsiao55 The thinking is that the gallery block is owned by triniti and cms-core doensn't override it, so the clean place to source the selected galleries `launch_text` and pass to poster preview is shared triniti layer. Then rendering would still live in cms-core and we wouldn't have to duplicate the gallery block modal to pass one prop.
> 
> @guizzettic you can override that gallery block in cms-core as well. We probably don't have that file in cms-core cause there was no need to change anything triniti was already doing. To me it feels weird to be adding something that's not going to be used. If we are going to be adding this prop in triniti, I would render it in triniti as well

@bhsiao55 Ok that's fair, I'll close this and handle it in cms-core.

bhsiao55:
> > > > @guizzettic I see you are adding the launchtext at the triniti level but you are not really rendering the text here. I assume you will be rendering it in cms-core but just wondering why you decided to add the launchtext here and not just do everything in cms-core?
> > > 
> > > 
> > > @bhsiao55 The thinking is that the gallery block is owned by triniti and cms-core doensn't override it, so the clean place to source the selected galleries `launch_text` and pass to poster preview is shared triniti layer. Then rendering would still live in cms-core and we wouldn't have to duplicate the gallery block modal to pass one prop.
> > 
> > 
> > @guizzettic you can override that gallery block in cms-core as well. We probably don't have that file in cms-core cause there was no need to change anything triniti was already doing. To me it feels weird to be adding something that's not going to be used. If we are going to be adding this prop in triniti, I would render it in triniti as well
> 
> @bhsiao55 Ok that's fair, I'll close this and handle it in cms-core.

@guizzettic you can still add it to triniti but I would add everything you need to render the launch text without requiring cms-core, unless somebody mentioned we didn't want this launch text feature in triniti and only tmz cms needs it

guizzettic:
> > > > > @guizzettic I see you are adding the launchtext at the triniti level but you are not really rendering the text here. I assume you will be rendering it in cms-core but just wondering why you decided to add the launchtext here and not just do everything in cms-core?
> > > > 
> > > > 
> > > > @bhsiao55 The thinking is that the gallery block is owned by triniti and cms-core doensn't override it, so the clean place to source the selected galleries `launch_text` and pass to poster preview is shared triniti layer. Then rendering would still live in cms-core and we wouldn't have to duplicate the gallery block modal to pass one prop.
> > > 
> > > 
> > > @guizzettic you can override that gallery block in cms-core as well. We probably don't have that file in cms-core cause there was no need to change anything triniti was already doing. To me it feels weird to be adding something that's not going to be used. If we are going to be adding this prop in triniti, I would render it in triniti as well
> > 
> > 
> > @bhsiao55 Ok that's fair, I'll close this and handle it in cms-core.
> 
> @guizzettic you can still add it to triniti but I would add everything you need to render the launch text without requiring cms-core, unless somebody mentioned we didn't want this launch text feature in triniti and only tmz cms needs it

@bhsiao55 Yeah that's true, but cms-core-js is already aliasing image-asset-picker-field, so anything rendered on triniti side gets replaced. I think handling it in cms-core-js makes more sense now.

bhsiao55:
> > > > > > @guizzettic I see you are adding the launchtext at the triniti level but you are not really rendering the text here. I assume you will be rendering it in cms-core but just wondering why you decided to add the launchtext here and not just do everything in cms-core?
> > > > > 
> > > > > 
> > > > > @bhsiao55 The thinking is that the gallery block is owned by triniti and cms-core doensn't override it, so the clean place to source the selected galleries `launch_text` and pass to poster preview is shared triniti layer. Then rendering would still live in cms-core and we wouldn't have to duplicate the gallery block modal to pass one prop.
> > > > 
> > > > 
> > > > @guizzettic you can override that gallery block in cms-core as well. We probably don't have that file in cms-core cause there was no need to change anything triniti was already doing. To me it feels weird to be adding something that's not going to be used. If we are going to be adding this prop in triniti, I would render it in triniti as well
> > > 
> > > 
> > > @bhsiao55 Ok that's fair, I'll close this and handle it in cms-core.
> > 
> > 
> > @guizzettic you can still add it to triniti but I would add everything you need to render the launch text without requiring cms-core, unless somebody mentioned we didn't want this launch text feature in triniti and only tmz cms needs it
> 
> @bhsiao55 Yeah that's true, but cms-core-js is already aliasing image-asset-picker-field, so anything rendered on triniti side gets replaced. I think handling it in cms-core-js makes more sense now.

@guizzettic yes that is true for our case but if somebody was not using our cms-core-js they would also get the launch text if you added to triniti

guizzettic:
@bhsiao55 After discussing keeping cms-js self sufficient so overlay is supported.

src/plugins/dam/components/image-asset-picker-field/index.js
```diff
@@ -4,14 +4,19 @@ import AssetPickerField from '@triniti/cms/plugins/dam/components/asset-picker-f
...
   const downloadUrl = damUrl(assetRef);
   const previewUrl = damUrl(assetRef, '1by1', 'sm');
 
   return (
     <div className="d-block mb-3">
-      <a href={downloadUrl} target="_blank" rel="noopener noreferrer" className="hover-box-shadow d-inline-block rounded-3">
+      <a href={downloadUrl} target="_blank" rel="noopener noreferrer" className="hover-box-shadow d-inline-block rounded-3 overflow-hidden position-relative">
         <Media src={previewUrl} alt="" width={200} height={200} object className="rounded-3" />
+        {launchText && (
+          <div className="position-absolute bottom-0 start-0 end-0 px-2 py-1 text-white text-break bg-black bg-opacity-75 lh-sm">
```
bhsiao55: @guizzettic I believe the launch text is always displayed with all caps on website.  I think it should display the same way

bhsiao55 (changes requested):

src/plugins/dam/components/image-asset-picker-field/index.js
guizzettic: @bhsiao55 Good catch, I updated it now 👍

bhsiao55 (approved):
lgtm

PR #545, opened 2026-05-19
Support Liveblog Posting

guizzettic:
* Add CMS support for live blog timelines and live blog update teasers.


src/plugins/curator/components/create-teaser-modal/LiveBlogUpdateTeaserModal.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { TextField } from '@triniti/cms/components/index.js';
+import TimelinePickerField from '@triniti/cms/plugins/curator/components/timeline-picker-field/index.js';
+import withTeaserModal from '@triniti/cms/plugins/curator/components/create-teaser-modal/withTeaserModal.js';
+
+function LiveBlogUpdateTeaserModal() {
+  return (
+    <>
+      <TextField name="title" label="Title" required />
+      <TimelinePickerField name="timeline_ref" label="Timeline" required />
```
rsumilang, line 10: @efjacobson should this be filtered to live blogs only?

rsumilang (changes requested):

src/plugins/curator/components/create-teaser-modal/LiveBlogUpdateTeaserModal.js
efjacobson: @rsumilang you mean, should the timeline picker be filtered to only allow timelines that have is_live_blog:true?

i don't think so - i expect it will be very obvious which ones are and are not live, and the constraint would add some ordering complexity for their workflow. we can always add it later but i don't think it's necessary right now. product stakeholders brought made the same suggestion but are fine with not filtering for mvp

rsumilang: @efjacobson if a wrong type of timeline is chosen, then it might not break anything in the cms but maybe on the web rendering if we don't have guardrails just to make sure the associated timelines are live blogs. just some thoughts.

efjacobson: @rsumilang yeah we could do that, as long as it is kept local to the teasers. (ie a liveblog teaser would refuse to render if not in a live timeline, as opposed to the timelines enforcing what kind of teaser they would render)

rsumilang (approved):

PR #548, opened 2026-06-17
Batch upload "Apply to All" no longer erases unsaved display title an…

rsumilang:
…d alt text on the active asset form.

### DIFF
--- a/src/components/with-form/index.js
+++ b/src/components/with-form/index.js
@@ -77,9 +77,7 @@ export default function withForm(Component, config = {}) {
               if (delegate.reinitialize) {
                 delegate.shouldReinitialize = false;
                 delegate.reinitialize = false;
-                const keepDirty = delegate.keepDirtyOnReinitialize !== undefined
-                  ? delegate.keepDirtyOnReinitialize
-                  : config.keepDirtyOnReinitialize;
+                const keepDirty = delegate.keepDirtyOnReinitialize ?? config.keepDirtyOnReinitialize;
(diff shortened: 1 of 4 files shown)


src/components/with-form/index.js
```diff
@@ -77,9 +77,13 @@ export default function withForm(Component, config = {}) {
               if (delegate.reinitialize) {
                 delegate.shouldReinitialize = false;
                 delegate.reinitialize = false;
+                const keepDirty = delegate.keepDirtyOnReinitialize !== undefined
```
hunyoboy: @rsumilang nit: can be - `const keepDirty = delegate.keepDirtyOnReinitialize ?? config.keepDirtyOnReinitialize`

src/components/with-form/index.js
```diff
@@ -77,9 +77,13 @@ export default function withForm(Component, config = {}) {
...
+                  ? delegate.keepDirtyOnReinitialize
+                  : config.keepDirtyOnReinitialize;
+                delete delegate.keepDirtyOnReinitialize;
+                form.setConfig('keepDirtyOnReinitialize', keepDirty);
```
hunyoboy: @rsumilang this works and fixes issue. but just curious if setting config here is needed:

1. the issue is fixed by just setting `keepDirtyOnReinitialize` to true inside `delegate.handleApplyToAll`.
2. it seems redundant since we call `form.setConfig` again inside setTimeout.

hunyoboy (approved):

PR #549, opened 2026-06-17
Updates to apple new modal

normtronics:
The Create Notification modal only lets the user pick the article.
Once the notification is created, the user can customize the notification text and choose the "Send At" time from the notification's detail/edit view.
The "Send At" field is removed from the modal entirely.

### DIFF
--- a/src/plugins/notify/components/create-notification-modal/withNotificationModal.js
+++ b/src/plugins/notify/components/create-notification-modal/withNotificationModal.js
@@ -44,16 +44,14 @@ export default function withNotificationModal(ModalFields, modalOptions = {}) {
         const content = getContent(values.content_ref);
         if (content) {
           values.title = content.get('title');
-        } else {
+        }
+
+        if (hideSendOptions || !content) {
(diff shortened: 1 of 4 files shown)


src/plugins/notify/components/create-notification-modal/withNotificationModal.js
```diff
@@ -44,10 +48,14 @@ export default function withNotificationModal(ModalFields) {
           values.send_on_publish = false;
         }
 
-        if (values.send_on_publish) {
+        if (values.send_on_publish || hideSendOptions) {
           values.send_at = null;
         }
 
+        if (hideSendOptions) {
+          values.send_on_publish = false;
```
rsumilang: When hideSendOptions=true AND content is null, send_on_publish is set to false twice. Not a bug, but redundant. This can be cleaned up a little for clarity.

rsumilang (changes requested):

rsumilang (approved):

